PR #9, opened 2022-03-17
Add missing documentation for DualView and ScatterView constructors

Rombur:


API-Reference/Kokkos::DualView.md
```diff
@@ -138,6 +138,30 @@ class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
...
+  ///                                                                                                                                                                                                                                        
+  /// This constructor works like the analogous constructor of View.                                                                                                                                                                         
+  /// The first arguments are wrapped up in a ViewCtor class, this allows                                                                                                                                                                    
+  /// for a label, without initializing, and all of the other things that can                                                                                                                                                                
+  /// be wrapped up in a Ctor class.
```
dalg24, line 149: What is "Ctor class"?

crtrott (approved):
We can do this, but the DualView documentation is not at all in the desired format. In the API Reference we don't document functions in the synopsis itself. But that rework could be a different step.

PR #17, opened 2022-05-02
algorithms: fix name and sidebar of std sorting

fnrizzi:
Fix inconsistency with std sorting algorithms page name and location


_Sidebar.md
```diff
@@ -15,15 +15,15 @@
...
-4. **[[Std Algorithms|StdAlgorithms]]**
+1. **[[Random Numbers|Random-Number-Generators]]**
+2. **[[Iterators|Iterators]]**
+3. **[[Std Algorithms|StdAlgorithms]]**
    1. **[[NonModSequenceOps|StdNonModSeq]]**
    2. **[[ModSeqOps|StdModSeq]]**
    3. **[[PartioningOps|StdPartioningOps]]**
    4. **[[Numeric|StdNumeric]]**
    5. **[[StdMinMaxElement|StdMinMaxElement]]**
+   6. **[[Sorting|StdSorting]]**
```
dalg24, line 26: Why do you think it should be moved there?
It could introduce some confusion such as:  Why is `Kokkos::sort` not in `Experimental::` like the rest of the algorithms?  Why does it not have overloads that take iterators like all other algorithms?

dalg24: Nevermind these are `Kokkos::Experimental::is_sorted[_until]`

dalg24 (approved):

PR #19, opened 2022-05-16
Add documentation on all Non modifying algorithms.

rgayatri23:
The PR adds basic information and usage of all non modifying algorithms.


StdNonModSeq.md
```diff
@@ -107,7 +107,7 @@ Returns an iterator to the *first* element in `[first, last)` for which the pred
 
 ### Parameters and Requirements
 
-- `exespace`, `first, last`, `view`: same as in `find` (TODO: link)
+- `exespace`, `first, last`, `view`: same as in [Kokkos::Experimental::find](#Kokkos::Experimental::find)
```
dalg24, line 110: Use code fonts 
```suggestion
- `exespace`, `first, last`, `view`: same as in [`Kokkos::Experimental::find`](#Kokkos::Experimental::find)
```
```suggestion
- `exespace`, `first, last`, `view`: same as in [Kokkos::Experimental::find](#Kokkos::Experimental::find)
```

StdNonModSeq.md
```diff
@@ -128,10 +128,9 @@ Returns an iterator to the *first* element in `[first, last)` for which the pred
   };
   ```
 
-
 ### Return
 
-- (1,2): `InputIterator` instance pointing to the first element satisfying condition, or `last` if no elements is found
+- (1,2): `InputIterator` instance pointing to the first element where the predicate is evaluated to true, or `last` if no element is found
```
dalg24, line 133: What was wrong with the original version?  I would have just done "elements -> such element"

StdNonModSeq.md
```diff
@@ -150,17 +149,999 @@ struct EqualsValue
...
-using view_type = Kokkos::View<int*>;
+auto exespace = Kokkos::DefaultExecutionSpace;
+using view_type = Kokkos::View<execspace, int*>;
 view_type a("a", 15);
 // fill a somehow
 
 // create predicate
 EqualsValue<int> p(5);
 
-auto exespace = Kokkos::DefaultExecutionSpace;
```
dalg24, line 160: FWIIW I liked better the original version

dalg24, line 160: FWIIW I liked better the original version

StdNonModSeq.md
```diff
@@ -150,17 +150,999 @@ struct EqualsValue
...
+Kokkos::pair<int,int> mismatch_index = KE::mismatch(Kokkos::OpenMP(), KE::begin(a), KE::end(a), KE::begin(b), KE::end(b), p);
+
+```
+
+
+## Kokkos::Experimental::equal
+```cpp
+template <class ExecutionSpace, class IteratorType1, class IteratorType2>
+bool equal(const ExecutionSpace& ex, IteratorType1 first1, IteratorType1 last1,                                             (1)
+           IteratorType2 first2);
```
fnrizzi, line 466: I would move the numbers closer to the left otherwise it will impact the rendering of the page

StdNonModSeq.md
```diff
@@ -150,17 +150,999 @@ struct EqualsValue
...
+
+```
+
+## Kokkos::Experimental::count_if
+
+```cpp
+template <class ExecutionSpace, class IteratorType, class Predicate>
+typename IteratorType::difference_type count_if(const ExecutionSpace& ex,
+                                                IteratorType first,
+                                                IteratorType last,                                             (1)
```
fnrizzi, line 570: same comment about the numbering (1) ... move closer to declaration

rgayatri23: Thanks. I aligned the numbers and moved them closer.

dalg24 (changes requested):
%s/Expeimental/Expreimental/g

fnrizzi:
please update this so that you use the new sphinx branch

rgayatri23:
> please update this so that you use the new sphinx branch. Keep in mind all editable sources from now on are under `source/` (the wiki-md-files-to-port directory will be deleted soon).

Done. Also changed the base to the new sphinx branch.

fnrizzi:
why is the md file in wiki-md-files-to-port changed?

rgayatri23:
> 

Not really sure. I did not commit that one. I see that the wiki-md-file is exactly the same as the update file that I committed. Is that a backup of the files that happens automatically ?

fnrizzi:
> > 
> 
> Not really sure. I did not commit that one. I see that the wiki-md-file is exactly the same as the update file that I committed. Is that a backup of the files that happens automatically ?

don't know, can you try to fix it somehow?

fnrizzi:
@rgayatri23 do you have any updates on this? We should try to finalize this soon ideally, because I would like to start updating things to account also the team level API too but first this needs to go in

rgayatri23:
@fnrizzi . I will finish it this week.

rgayatri23:
This is being continued in #170

fnrizzi:
closing this since we have 170

PR #24, opened 2022-05-18
Issue/5017

ajpowelsnl:
This change addresses #5017 by adding `Kokkos::fence()` after the end of the `parallel_for()` bloc in the UVM example in `ProgrammingGuide/Kokkos-and-Virtual-Functions.md.`


masterleinad:
5 commits for a one-line change. :rofl: Have you tried that this is actually necessary (and works)?

ajpowelsnl:
Do over

PR #26, opened 2022-05-19
Data-Parallelism.md: rework explanation of tags

ajpowelsnl:
Reworked explanation of `parallel_*` tags for clarity

### DIFF
--- a/API-Reference/Data-Parallelism.md
+++ b/API-Reference/Data-Parallelism.md
@@ -30,7 +30,7 @@ Reducer objects used in conjunction with [parallel_reduce](Kokkos%3A%3Aparallel_
 
 ## Tags for Team Policy Calculations
 
-Team size (`team_size_max`, `team_size_recommended`) is determined by the particular parallel patttern tag.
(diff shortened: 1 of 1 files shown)


JBludau (commented):
Typo in pattern.

API-Reference/Data-Parallelism.md
```diff
@@ -30,7 +30,7 @@ Reducer objects used in conjunction with [parallel_reduce](Kokkos%3A%3Aparallel_
 
 ## Tags for Team Policy Calculations
 
-Tags passed to `team_size_max` or `team_size_recommended` to identify whether the pattern for which the team size for a functor is being requested
+Team size (`team_size_max`, `team_size_recommended`) is determined by the particular parallel patttern tag.
```
masterleinad, line 33: The team size is not directly determined by the tag but the tags rather tell you which overload is requested.

JBludau: hmm I think if it should be as crisp of a description as possible, we could stick to Damien and just get rid of the whether in the original

masterleinad (approved):

PR #27, opened 2022-05-21
Improve macros documentation

cz4rs:
May 16, 2022 Documentation Bootcamp


API-Reference/Macros.md
```diff
@@ -76,25 +76,81 @@ with.
...
+| ------| ---------------|
+| `KOKKOS_COMPILER_NVCC` | Defined if NVIDIA compiler is used. |
+| `KOKKOS_COMPILER_GNU`  | Defined if GCC compiler is used. |
+| `KOKKOS_COMPILER_INTEL` | Defined if Intel compiler or Intel LLVM compiler is used. |
+| `KOKKOS_COMPILER_IBM`   | Defined if IBM compiler or XLClang is used. |
+| `KOKKOS_COMPILER_CRAYC` | Defined if CRAY compiler is used for host code. |
+| `KOKKOS_COMPILER_APPLECC` | Defined if Apple cc compiler is used. |
+| `KOKKOS_COMPILER_CLANG` | Defined if Clang compiler is used. |
+| `KOKKOS_COMPILER_PGI` | Defined if PGI compiler is used. |
+| `KOKKOS_COMPILER_MSVC` | Defined if MSVC compiler is used. |
```
masterleinad, line 93: What about `KOKKOS_COMPILER_NVHPC`?

API-Reference/Macros.md
```diff
@@ -76,25 +76,81 @@ with.
...
 
+## Compilers
+
+The following macros can be used to query the integral version values of used compilers.
+
+| Macro | Description |
+| ------| ---------------|
+| `KOKKOS_COMPILER_NVCC` | Defined if NVIDIA compiler is used. |
+| `KOKKOS_COMPILER_GNU`  | Defined if GCC compiler is used. |
+| `KOKKOS_COMPILER_INTEL` | Defined if Intel compiler or Intel LLVM compiler is used. |
```
masterleinad, line 87: I think `Intel` calls them nowadays `Intel Compiler Classic` and `Intel oneAPI DPC++/C++ Compiler`, see https://www.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top.html and https://www.intel.com/content/www/us/en/develop/documentation/oneapi-dpcpp-cpp-compiler-dev-guide-and-reference/top.html.

API-Reference/Macros.md
```diff
@@ -76,25 +76,81 @@ with.
...
+
+## Compiler Extension for Atomics
+
+| Macro | Description |
+| ------| ---------------|
+| `KOKKOS_ENABLE_CUDA_ATOMICS` | Defined if CUDA atomics is enabled when compiling NVIDIA device code. |
+| `KOKKOS_ENABLE_GNU_ATOMICS` | Defined if GCC atomics is enabled when compiling host code. |
+| `KOKKOS_ENABLE_INTEL_ATOMICS` | Defined if Intel atomics is enabled. |
+| `KOKKOS_ENABLE_OPENMP_ATOMICS` | Defined if OpenMP atomics is enabled. |
+| `KOKKOS_ENABLE_WINDOWS_ATOMICS` | Defined if Win32 atomics is enabled. |
```
masterleinad, line 103: I would think that these are meant to be internal.

API-Reference/Macros.md
```diff
@@ -76,25 +76,81 @@ with.
...
+| `KOKKOS_ENABLE_PRAGMA_UNROLL` | Defined if the compiler is enabled to do the loop unrolling. [#pragma unroll](https://www.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/pragmas/intel-specific-pragma-reference/unroll-nounroll.html) |
+| `KOKKOS_ENABLE_PRAGMA_VECTOR` | Defined if the compiler is enabled to force vectorization. [#pragma vector](https://www.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/pragmas/intel-specific-pragma-reference/vector-1.html) |
+
+## Allocation Mechanism
+
+| Macro | Description |
+| ------| ---------------|
+| `KOKKOS_ENABLE_INTEL_MM_ALLOC` | Defined if set to use Intel [_mm_malloc](https://www.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/intrinsics/data-align-mem-alloc-intrins-and-inline-asmbly/allocating-and-freeing-aligned-memory-blocks.html) for the HostSpace memory allocation. |
```
masterleinad, line 119: Again, I would think this is supposed to be internal.

ProgrammingGuide/Compiling.md
```diff
@@ -338,6 +338,7 @@ The following options control `find_package` paths for CMake-based TPLs:
 
 #### Architecture Keywords
 Architecture-specific optimizations can be enabled by specifying `-DKokkos_ARCH_X`.
+See the [complete list of architecture keywords](../API-Reference/Macros.md#architectures).
 
 * Kokkos_ARCH_A64FX
```
cz4rs, line 343: Should we consider removing the list below (as duplicated) and only leave the link to `API-Reference/Macros.md`?

ldh4: The only concern I have about removing the list below is that there is a very subtle capitalizations differences between the keywords listed on this page and the ones in `API-Reference/Macros.md`.
Here, it is `Kokkos_ARCH_X` whereas in `Macros.md`, it's `KOKKOS_ARCH_X`.
Though for CMake builds, using either doesn't seem to make any difference...

masterleinad: I think it's debatable whether users should be able to rely on the existence of these macros.

The `CMake` option always looks like `Kokkos_ENABLE...` (although we also define local CMake variables `KOKKOS_ENABLE...`) while the macro is `KOKKOS_ENABLE...`.

crtrott: note I don't think that KOKKOS_ENABLE is valid from our side.

ajpowelsnl: @crtrott , @masterleinad -- Just to make sure I understand the principle here, you are proposing:
1) Users *should not* use Kokkos macros, that macros are/ should be mostly / exclusively for developers;
2) the CMake variables, `Kokkos_ENABLE_X` , `Kokkos_ARCH_XYZ` should be the only user-facing variables (vs. Macros  + CMake variables)?
3) CMake variables will take precedence, and if not set, the macro or environment variable will be used?

fnrizzi:
is this ready? if not, can you please make it draft?

cz4rs:
Converting to draft to fix conflicts and check recent changes.

PR #103, opened 2022-07-05
Document new realloc overloads taking Kokkos::view_alloc() return type

masterleinad:
Corresponding to https://github.com/kokkos/kokkos/pull/5035.


crtrott (commented):
I don't like the way this is done (e.g. exposing implementation details) but its what the existing one already did.

PR #104, opened 2022-07-05
Document new resize overloads taking Kokkos::view_alloc() return type

masterleinad:
Corresponding to https://github.com/kokkos/kokkos/pull/4844.

### DIFF
--- a/docs/source/API/core/view/resize.md
+++ b/docs/source/API/core/view/resize.md
@@ -35,12 +35,31 @@ void resize(const I& arg_prop, Kokkos::View<T, P...>& v,
        const size_t n6 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
        const size_t n7 = KOKKOS_IMPL_CTOR_DEFAULT_ARG);
 
+template <class T, class... P, class... ViewCtorArgs>
+void resize(const Impl::ViewCtorProp<ViewCtorArgs...>& arg_prop,
+       Kokkos::View<T, P...>& v,
+       const size_t n0 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+       const size_t n1 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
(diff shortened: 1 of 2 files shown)


docs/source/API/core/view/resize.md
```diff
@@ -104,6 +145,20 @@ void resize(const I& arg_prop, Kokkos::View<T, P...>& v,
   * `layout`: a layout instance containing the new dimensions.
   * `arg_prop`: View constructor property, e.g., `Kokkos::WithoutInitializing`.
 
+* ```c++
+  template <class T, class... P, class... ViewCtorArgs>
+  void resize(const Impl::ViewCtorProp<ViewCtorArgs...>& arg_prop,
+         Kokkos::View<T, P...>& v,
+         const typename Kokkos::View<T, P...>::array_layout& layout);
+  ```
+  Resizes `v` to have the new dimensions while preserving the contents for the common subview of the old and new view. The new `Kokkos::View` is constructed using the View constructor properties `arg_prop`, e.g., `Kokkos::view_alloc(Kokkos::WithoutInitializing)`.
```
dalg24, line 154: What about use of execution space if provided to copy elements

dalg24 (commented):
Did we discuss anywhere whether we really want to support long term both single argument and bundle of them via `view_alloc`?

docs/source/API/core/view/resize.md
masterleinad: I added a sentence to both overloads. What I really want to say, though, is that providing an execution space follows standard semantics, i.e., use it as much as possible internally and enqueue operations without explicitly fencing. I'm not quite sure if that needs extra documentation and if so where a good place at a central location for such a statement would be.

crtrott (approved):
I don't like how some of this exposes implementation details, but a lot of it already did. So lets merge it and then do a systematic cleanup later.

fnrizzi:
@masterleinad can you please rebase/resync? I will merge it afterward

masterleinad:
> @masterleinad can you please rebase/resync? I will merge it afterward

Rebased.

Issue #33, opened 2022-06-06
sphinx porting: things to add 


fnrizzi:
- more visible link to GH page (maybe repo icon at the top?) 
   - [x] propose a few ways to do this 
   - [x] we pick one 
   - [x] finalize 

- [x] we need to make sure each doc page when rendered has a visible link to its rst or md file int the source repo so that if one notices typos or things, one can easily click on the link and be taken to edit that page directly
   - [x] figure out ways to do this and propose 
   - [x] we pick one 
   - [x] finalize

marcinwrobel1986:
Regarding: 
> more visible link to GH page

I would put the link to the documentation at the top of main [README.md](https://github.com/kokkos/kokkos#readme) as it is at the moment. Like marked below:
![Zrzut ekranu z 2022-06-27 14-54-03](https://user-images.githubusercontent.com/47597569/175948321-36fc5a3f-0c21-4e0b-98af-583210b4cdfe.png)

marcinwrobel1986:
Regarding:
> we need to make sure each doc page when rendered has a visible link to its rst or md file int the source repo so that if one notices typos or things, one can easily click on the link and be taken to edit that page directly

1. Localization of files in the repository must be FROZEN
  - when it's not frozen, then links to the code will be broken, each time files are migrated somewhere
  - best would be to agree when documentation is deployed? Once it's merged to develop? Or maybe only  when it's been merged to master branch
  - once there is a consensus, regarding which branch should links point to, these links for editing files could be created
2. Once the files localization is frozen and the branch to which links should point is chosen, then the process of links creation could be defined
  - should it be done by hand? Potential problems:
    - humans makes mistakes
    - one needs to remember to put a link when adding some site in the wiki
  - should it be automated? Possible variants:
    - links is added to the *.md or *.rst file
    - links are added to generated *.html files
  - depending on chosen way (when links are added) a automation script must be written (preferable Python3 one)
3. Regardless of which way of links creation will be chosen, the place where link appears in the wiki page must be defined as well as it's appearance.

SUMMARY:
- location must be final
- I would personally suggest to automate this job
- IMHO the best would be to edit html's after generation, why? Script always have to do the same to html, right? There would be no distinguish between how to put a link in *.md or *.rst file. Adding links to different format probably could be done without a problem, but it adds another level of complexity and it's error prone.

marcinwrobel1986:
Done with https://github.com/kokkos/kokkos-core-wiki/pull/109
[...]

PR #109, opened 2022-07-12
#33: Add edit button which redirects to GitHub

marcinwrobel1986:
### THIS PR:
- added `edit_button_handler.py` script for adding `edit` icon to generated html pages
- added step to `deploy_docs.yml` (GitHub action)
- added `edit-button` section to make file in order to simplify edit button generation locally

@dalg24 @fnrizzi Please take a look. 

Documentation DEMO with working button on the bottom right could be found [here](https://marcinwrobel1986.github.io/kokkos-core-wiki/)

Below are sample screenshots (2nd with mouse over edit button):
Zrzut ekranu z 2022-07-15 10-20-48
Zrzut ekranu z 2022-07-15 10-21-02


fnrizzi:
@marcinwrobel1986 this look good, i would have these suggestions though: 
- can you please show how the button would look like if placed at the top right of the page instead of bottom? 
- can you please change the button link to take to the source file 
- can you please change he text "edit on github" to "take me to souce file" 

I think we need to avoid editing files directly on github

marcinwrobel1986:
> @marcinwrobel1986 this look good, i would have these suggestions though:
> 
> * can you please show how the button would look like if placed at the top right of the page instead of bottom?

IMHO the place on the bottom right is the only one which will not interfered with the theme. Current position is the same for every page

> * can you please change the button link to take to the source file

Done.

> * can you please change he text "edit on github" to "take me to source file" ? I think we need to avoid editing files directly on github.

Done as well.

dalg24:
I pulled your branch and built it locally and cannot see the button.

dalg24:
I also would slightly prefer the `Edit on GitHub` to be in the header, just like with the ReadTheDocs theme.

dalg24:
> you need to run Python script python3 source/edit_button_handler.py

We need to add to the Makefile too

marcinwrobel1986:
> > you need to run Python script python3 source/edit_button_handler.py
> 
> ~We need to add to the Makefile too~ Actually I would say it needs to be added to the html rule in the Makefile and nowhere else

Hey @dalg24 Damien,
I implemented this more elegant version with edit button at the top, also Makefile is modified and used in CI.

fnrizzi (approved):

fnrizzi:
i think this looks nice ! maybe to make it more visible we could colorcode it? just a nitpick comment

docs/Makefile
```diff
@@ -3,3 +3,6 @@ html:
 
 clean:
 	rm -rf ./generated_docs/
+
+edit-button:
+	python3 ./source/edit_button_handler.py
```
dalg24, line 8:
```suggestion
	python3 ./source/edit_button_handler.py

all:
	make html
	make edit-button
```
Or 
```
html:
	sphinx-build -b html ./source/ ./generated_docs/
	python3 ./source/edit_button_handler.py

marcinwrobel1986: @dalg24 done

dalg24 (approved):

PR #122, opened 2022-07-28
#119: Fix edit button duplication

### LINKED ISSUE #119: Edit button is being duplicated every time the documentation is rebuilt
I can live with it but I figured it might be something that is not too hard to fix.
All pages that are not modified get an extra edit button every time the documentation is re-built.

marcinwrobel1986:
### THIS PR:
- removed anything related to image from previous concept
- added check if the edit button is already added to the HTML file

closes #119


dalg24 (approved):
Thanks for fixing

PR #125, opened 2022-07-29
Add DefaultExecutionSpace to API

JBludau:
I noticed, that the `Default(Host)ExecutionSpace` only exists in examples or in the programming guide. As part of the public interface it should probably have a API documentation. 
I propose a new section because of the alias' dependency on the configuration.


dalg24 (commented):
I just wanted to mention the (IMO misplaced) information about these here
https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/Initialization.html#initialization-by-command-line-arguments

crtrott:
I think the whole order is messed up but otherwise ok.

crtrott (approved):

PR #126, opened 2022-07-29
Adding Hip spaces ... already without `Experimental`

JBludau:
It contains only a minimal hint on the `xnack` rabbithole


dalg24 (changes requested):
Add the `Experimental::`.  We will need to add some language about when the promotion to the `Kokkos::` namespace happened.

JBludau:
how about this?
image

dalg24 (approved):

Rombur (approved):

PR #128, opened 2022-07-29
Remove copy from ConstBitset to ConstBitset

Azrael3000:
This does not work in the code (as expected) and thus should not be listed here.

@dalg24 as discussed briefly on Slack today.


dalg24 (approved):
Uggh...  Thanks for fixing it

PR #129, opened 2022-07-29
Add Kokkos_ARCH_NATIVE to cmake flags documentation

JBludau:


docs/source/keywords.rst
```diff
@@ -272,6 +272,10 @@ Architecture Keywords
       - Description/info
       - Default
 
+    * * ``Kokkos_ARCH_NATIVE``
+      * Let compiler detect the architecture of the machine compiling on
```
masterleinad, line 276:
```suggestion
      * Optimize for the local CPU architecture
```
Would be more inline with with the way other architectures in `BUILD.md` are treated here.

PR #133, opened 2022-08-01
Fix last RangePolicy example

brian-kelley:
``parallel_for(RangePolicy<>(N), ...)`` doesn't quite work because RangePolicy's constructor requires start,end (not just end).
Replace it with ``parallel_for(RangePolicy<>(0, N), ...)``.


nmm0 (approved):
Looks good to me -- if you get a chance could you fix the ones on lines 104 and 107 that are also outdated?

brian-kelley:
@nmm0 Just added. Didn't notice those before

PR #135, opened 2022-08-01
warn in view that pointer manipulation might end in undefined behavior

JBludau:
This is part of the general discussion of making all user facing memory coarse grained and what happens if the user gets the pointer and changes how the os handles the underlying memory. 

I am not sure about the wording though...maybe it is better to say:
WARNING: calling any function that manipulates the behavior of the memory (e.g. `memAdvise`) on memory managed by `Kokkos` is illegal.

### DIFF
--- a/docs/source/API/core/c_style_memory_management/malloc.md
+++ b/docs/source/API/core/c_style_memory_management/malloc.md
@@ -18,6 +18,8 @@ If allocation succeeds, returns a pointer to the lowest (first) byte in the allo
 
 If allocation fails, an exception of type `Kokkos::Experimental::RawMemoryAllocationFailure` is thrown.
 
+WARNING: calling any function that manipulates the behavior of the memory (e.g. `memAdvise`) on memory managed by `Kokkos` results in undefined behavior. You are on your own with this.
+
 ## Parameters
 
 `label`: A user provided string which is used in profiling and debugging tools via the KokkosP Profiling Tools.  
--- a/docs/source/API/core/c_style_memory_management/malloc.md
+++ b/docs/source/API/core/c_style_memory_management/malloc.md
@@ -18,7 +18,7 @@ If allocation succeeds, returns a pointer to the lowest (first) byte in the allo
 
 If allocation fails, an exception of type `Kokkos::Experimental::RawMemoryAllocationFailure` is thrown.
 
-WARNING: calling any function that manipulates the behavior of the memory (e.g. `memAdvise`) on memory managed by `Kokkos` results in undefined behavior. You are on your own with this.
+WARNING: calling any function that manipulates the behavior of the memory (e.g. `memAdvise`) on memory managed by `Kokkos` results in undefined behavior.
 
 ## Parameters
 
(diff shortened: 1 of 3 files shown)


PhilMiller:
Wording-wise, it's probably best to just match the C++ standard, and say "Doing X results in undefined behavior". That explicitly disclaims any promise of both what results, and whether the code diagnoses it in any given version, platform, or configuration.

JBludau:
Just to record it here:
The user would be fine calling whatever he wants on the memory managed by Kokkos for his use outside of our interface. But in this case he would have to make sure, that the annotations on the memory are reset to what we expect before using it with out API.
 We set the configuration we want at allocation time and do not check them when the view is used but we also do not do a good job in communicating the annotations we use on our memory.  
 
 Given this, it is actually not undefined behavior ... but I am not sure if we want to document "you can do advise, just make sure you revert it to the original state before using it with our API" or if we just say it is illegal/undefined behavior

docs/source/API/core/c_style_memory_management/malloc.md
```diff
@@ -18,6 +18,8 @@ If allocation succeeds, returns a pointer to the lowest (first) byte in the allo
 
 If allocation fails, an exception of type `Kokkos::Experimental::RawMemoryAllocationFailure` is thrown.
 
+WARNING: calling any function that manipulates the behavior of the memory (e.g. `memAdvise`) on memory managed by `Kokkos` results in undefined behavior. You are on your own with this.
+
```
ajpowelsnl, line 22: I would not include the second sentence, because the first communicates this fact.

docs/source/API/core/c_style_memory_management/realloc.md
```diff
@@ -9,6 +9,8 @@ void* kokkos_realloc(void* ptr, size_t new_size);
 
 Reallocates the given area of memory. It must be previously allocated by [`Kokkos::kokkos_malloc()`](malloc) or [`Kokkos::kokkos_realloc()`](realloc) on the same memory space [`MemorySpace`](../memory_spaces) and not yet freed with [`Kokkos::kokkos_free()`](free), otherwise, the results are undefined.
 
+WARNING: calling any function that manipulates the behavior of the memory (e.g. `memAdvise`) on memory managed by `Kokkos` results in undefined behavior. You are on your own with this.
+
 ## Parameters
```
ajpowelsnl, line 14: Same comment as above.

docs/source/API/core/view/view.md
```diff
@@ -212,7 +212,7 @@ Template parameters other than `DataType` are optional, but ordering is enforced
   * ```c++
     constexpr pointer_type data() const
     ```
-    Return the pointer to the underlying data allocation.
+    Return the pointer to the underlying data allocation. WARNING: calling any function that manipulates the behavior of the memory (e.g. `memAdvise`) on memory managed by `Kokkos` results in undefined behavior. You are on your own with this.
   * ```c++
```
ajpowelsnl, line 216: Leave out the second sentence please.

ajpowelsnl (changes requested):
Looks good, but please get rid of the second sentence.  It sounds a bit antagonistic, and doesn't convey any additional meaning beyond that of the first.

docs/source/API/core/c_style_memory_management/malloc.md
JBludau: done

docs/source/API/core/c_style_memory_management/realloc.md
JBludau: done

docs/source/API/core/view/view.md
JBludau: done

ajpowelsnl (approved):
Looks good!

crtrott (approved):

Issue #120, opened 2022-07-27
sphinx porting stage2: update API pages to use sphinx syntax 


fnrizzi:
- make a list of the API pages/features that need to be updated in priority order
  1. `View`
  2. all the ones inside [containers](https://kokkos.github.io/kokkos-core-wiki/API/containers-index.html)
  3. convert all parallel dispatch
  4. TBD
- after the list is approved by the team, make PR for each one to convert it from the "old" syntax to the "new" syntax 

By "new syntax" we mean the syntax that @nmm0 proposed to use, see [here](https://github.com/nmm0/kokkos-doc-test/blob/main/test_nm/source/api_reference/view.rst)

marcinwrobel1986:
## i. Highest priority:

| File name                                                        | Doc title | Done with PR |
|------------------------------------------------------------------| --------- | ------------ |
| `building.md`                                                    | [Build, Install and Use](https://kokkos.github.io/kokkos-core-wiki/building.html) | |
| `API/alphabetical.md`                                            | [API in Alphabetical Order](https://kokkos.github.io/kokkos-core-wiki/API/alphabetical.html) | |
| `ProgrammingGuide/Atomic-Operations.md`                          | [Atomic Operations](https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/Atomic-Operations.html) | |
| `ProgrammingGuide/Compiling.md`                                  | [Compiling](https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/Compiling.html) | |
| `ProgrammingGuide/HierarchicalParallelism.md`                    | [Hierarchical Parallelism](https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/HierarchicalParallelism.html) | |
| `ProgrammingGuide/Initialization.md`                             | [Initialization](https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/Initialization.html) | |
| `ProgrammingGuide/Interoperability.md`                           | [Interoperability and Legacy Codes](https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/Interoperability.html) | |
| `ProgrammingGuide/Introduction.md`                               | [Introduction](https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/Introduction.html) | |
[...]

PR #148, opened 2022-08-29
#120: API-core-view-view_like from md to rst

marcinwrobel1986:
### THIS PR:
- API/core/view/view_like transition from .md to .rst

Below render of the API/core/view/view_like after transition:

Zrzut ekranu z 2022-08-29 16-58-27


fnrizzi (approved):

dalg24:
Is there any change in rendering between the two?
I can't tell

marcinwrobel1986:
> Is there any change in rendering between the two? I can't tell

No :)
I thought you guys want to move from .md to .rst, but @fnrizzi told me this morning that it is valid only where there is C++ syntax, so now I will ignore such files, where there is no C++ syntax.

fnrizzi:
> > Is there any change in rendering between the two? I can't tell
> 
> No :) I thought you guys want to move from .md to .rst, but @fnrizzi told me this morning that it is valid only where there is C++ syntax, so now I will ignore such files, where there is no C++ syntax.

yes exactly

Issue #151, opened 2022-08-31
Reflect deprecation of need for volatile in reducer-interfacing bits


PhilMiller:
Reflect kokkos/kokkos#4077, kokkos/kokkos#4931, kokkos/kokkos#1554, kokkos/kokkos#5194, kokkos/kokkos#5215

ajpowelsnl:
Hi @PhilMiller -- do the PR mentioned above address main points of this issue?  If so, can we close the issue?  If not, can you detail what else needs to be done?

masterleinad:
Fixed by https://github.com/kokkos/kokkos-core-wiki/pull/152 and https://github.com/kokkos/kokkos-core-wiki/pull/185.

PR #152, opened 2022-08-31
#151: Clear out references to obviated volatile-qualified overloads

PhilMiller:
From 3.7 onwards, these are no longer necessary, and they are being deleted in the code base.

### DIFF
--- a/docs/source/API/core/stl-compat/pair.md
+++ b/docs/source/API/core/stl-compat/pair.md
@@ -29,15 +29,9 @@ struct pair {
   
     template <class U, class V>
     KOKKOS_FORCEINLINE_FUNCTION constexpr pair(const pair<U, V>& p);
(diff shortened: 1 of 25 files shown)


docs/source/API/core/stl-compat/pair.md
```diff
@@ -97,16 +83,8 @@ struct pair {
...
- 
-  * 
-  ```c++ 
-  template <class U, class V>
-  KOKKOS_FORCEINLINE_FUNCTION void operator=(const volatile pair<U, V>& p) volatile;
-  ```
-  
   Sets `first` to `p.first` and `second` to `p.second` 
-
+
```
masterleinad, line 87:
```suggestion

```

masterleinad (approved):

PR #156, opened 2022-09-01
Partitionspace

crtrott:


dalg24 (commented):
Are we supposed to remove the file from `wiki-md-files-to-port/`?

dalg24 (commented):
Please confirm that you did not change the file or say what you updated

crtrott:
I did not change the file.

dalg24 (approved):

dalg24:
> Are we supposed to remove the file from `wiki-md-files-to-port/`?

@crtrott @fnrizzi  do you know the answer to this?

fnrizzi:
I left the md-files directory because that is what we decided to do, but also said that it is not needed anymore since everything was ported and that at some point one would delete those files

PR #158, opened 2022-09-04
split std algorithms into separate files

fnrizzi:
The motivation of PR is that we had huge files with many algorithms all together. Soon we will add the team-level API and more examples, so this will make things very cluttered and hard to edit/find/browse etc.  

So this PR does *not change* the content, but implies these changes: 
- separate algorithms into individual files 
- ensure that each API snippet has is clearly within the ``Kokkos::Experimental`` namespace 

NOTE: the non-mod ops are not updated because we first have to merge [this other PR](#19 )


fnrizzi:
@crtrott @dalg24 I pinged you because even if nothing changes , it feels weird to merge it myself even if I could

dalg24:
> NOTE: the non-mod ops are not updated because we first have to merge [this other PR](https://github.com/kokkos/kokkos-core-wiki/pull/158#19)

Which PR are you referring to?

fnrizzi:
> > NOTE: the non-mod ops are not updated because we first have to merge [this other PR](https://github.com/kokkos/kokkos-core-wiki/pull/158#19)
> 
> Which PR are you referring to?

this PR :)

PR #161, opened 2022-09-07
Document promotion of math constants to `Kokkos::numbers::`

dalg24:
See kokkos/kokkos#5434

### DIFF
--- a/docs/source/API/core/numerics/mathematical-constants.md
+++ b/docs/source/API/core/numerics/mathematical-constants.md
@@ -1,11 +1,11 @@
 # Mathematical constants
 
 Defined in header [`<Kokkos_MathematicalConstants.hpp>`](https://github.com/kokkos/kokkos/blob/develop/core/src/Kokkos_MathematicalConstants.hpp)
-which is included from `<Kokkos_Core.hpp>` (since Kokkos 3.6)
+which is included from `<Kokkos_Core.hpp>`
 
 Provides all mathematical constants from [`<numbers>`](https://en.cppreference.com/w/cpp/numeric/constants) (since C++20).
 
-All constants are defined in the `Kokkos::Experimental` namespace.
+All constants are defined in the `Kokkos::numbers::` namespace since version 4.0, in `Kokkos::Experimental` in previous versions.
 
 **Mathematical constants**
 `e`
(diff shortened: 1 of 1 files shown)


docs/source/API/core/numerics/mathematical-constants.md
```diff
@@ -1,11 +1,11 @@
 # Mathematical constants
 
 Defined in header [`<Kokkos_MathematicalConstants.hpp>`](https://github.com/kokkos/kokkos/blob/develop/core/src/Kokkos_MathematicalConstants.hpp)
-which is included from `<Kokkos_Core.hpp>` (since Kokkos 3.6)
+which is included from `<Kokkos_Core.hpp>`
 
 Provides all mathematical constants from [`<numbers>`](https://en.cppreference.com/w/cpp/numeric/constants) (since C++20).
 
-All constants are defined in the `Kokkos::Experimental` namespace.
+All constants are defined in the `Kokkos::numbers::` namespace since version 4.0, in `Kokkos::Experimental` in previous versions.
```
fnrizzi, line 8: following up to what we said during meeting, I would suggest adding a shiny statement that justifies the choice of lowercase n

dalg24: Please suggest wording

fnrizzi: Here is a draft: 

NOTE: the namespace ``numbers`` is intentionally lower case. This is a rule we adopt in Kokkos to include code that is just simply a (back) port of C++ library functionalities in the corresponding namespace. In this specific case, ``Kokkos::numbers::`` includes a back port of C++20 functionalities. Whenever we will adopt C++20 as the minimum standard for Kokkos, you will thus have the choice of simply replace `Kokkos::numbers::` with ``std::numbers::``.

dalg24: To be honest I think this is a lot of text for something that users don't really need to know when they browse this page.

fnrizzi: i am totally fine putting that in a different page too but i think it is important to write *somewhere* why we pick a lowercase, if nothing else just to explicitly convey that this is an intentional choice

fnrizzi: i guess my point is that this breaks the kokkos convention so IMO we need to make sure users are aware this is not a slip or lack of care but there is a reason

dalg24: How can we make progress on it?  How about we merge as proposed here and we address https://github.com/kokkos/kokkos-core-wiki/pull/161#discussion_r965227309 elsewhere?

fnrizzi (approved):

PR #162, opened 2022-09-08
Extention of VirtualFunction with how to avoid V-Table lookup for pointers to derived

JBludau:
This is from a slack channel explanation we provided for a user question.
The question was about why calls to virtual functions from a pointer of derived type do fail on Rocm but not on nvcc.
https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1662487898753699


docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,60 @@ deviceInstance->setAField(someHostValue); // set some field on the host
 
 This is the solution that the code teams we have talked to have said is the most productive way to solve the problem.
 
+## But what if I do not really need the V-Tables on the device side?
+
+There are cases where a lot of the logic of a code is implemented using virtual functions. Nevertheless, the performance critical part of the code might not use parent pointers and dynamic polymorphism to compute a result from data. In these cases, the device might not need to have a working virtual function table. Consider the following example:
```
dalg24, line 116:
```suggestion
There are cases where a lot of the logic of a code is implemented using virtual functions. Nevertheless, the performance critical part of the code might not use pointers to base class and dynamic polymorphism to compute a result from data. In these cases, the device might not need to have a working virtual function table. Consider the following example:
```
First sentence seems unnecessary.

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,60 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+```c++
+struct Interface
+{
+    KOKKOS_DEFAULTED_FUNCTION
+    virtual ~Interface() = default;
+
+    KOKKOS_FUNCTION
+    virtual void operator()( const size_t, const Kokkos::View<int*> &) const = 0;
```
dalg24, line 124: You don't need the view to make your point.

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,60 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+{
+    ... 
+    Kokkos::View<int *> v("my view");
+    auto implementationPtr_h = std::make_shared<Implementation>();
+    implementationPtr_h->apply(v);
+}
+```
+### What is the problem?
+
+Inside the `parallel_for` the `operator()` is called which is maked `override`. On Rocm 5.2 this results in a memory access violation. When executing the `this->operator()(i,v)` call the runtime looks into the V-Table and dereferences a host pointer on the device.
```
masterleinad, line 155:
```suggestion
Inside the `parallel_for` the `operator()` is called which is marked `override`. On ROCm 5.2 this results in a memory access violation. When executing the `this->operator()(i,v)` call, the runtime looks into the V-Table and dereferences a host pointer on the device.
```

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,60 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+
+### But if that is the case, why does it work with NVCC?
+
+Notice, that the `parallel_for` is called from a pointer of type `Implementation` and not a pointer of type `Interface` pointing to an `Implementation` object. Thus, no V-Table lookup for the `operator()` would be necessary as it can be deduced from the context of the call that it will be the `operator()` in the class `Interface`. But here it comes down to how the compiler handles the lookup. NVCC uses that it sees the call coming from an `Implementation` object and thinks: "Oh, I see, that you are calling from a `Implementation` object, I know it will be the `operator()` in this class scope, I will do this for you". Rocm on the other hand sees your call and thinks “Oh, this is a call to a virtual method, I will look that up for you” - failing to read from the virtual function table, as it is containing host addresses.
```
masterleinad, line 159:
```suggestion
Notice, that the `parallel_for` is called from a pointer of type `Implementation` and not a pointer of type `Interface` pointing to an `Implementation` object. Thus, no V-Table lookup for the `operator()` would be necessary as it can be deduced from the context of the call that it will be the `operator()` in the class `Interface`. But here it comes down to how the compiler handles the lookup. NVCC understands that the call is coming from an `Implementation` object and thinks: "Oh, I see, that you are calling from an `Implementation` object, I know it will be the `operator()` in this class scope, I will do this for you". ROCm, on the other hand, sees your call and thinks “Oh, this is a call to a virtual method, I will look that up for you” - failing to read from the virtual function table, as it is containing host addresses.
```

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,60 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+
+### How to solve this?
+Stictly speaking the observed behavior on NVCC is an optimization that uses the context information to avoid the V-Table lookup. If the compiler does not apply this optimization you can help in different ways by providing additional information. 
+
+- Changing the `override` to `final` on the `operator()` in the `Implementation` class. This tells the compiler the `operator()` is not changing in derived objects and thus it knows which one to call without the V-Table. Nevertheless, this also prevents any overloads of `operator()` in classes derived from `Implementation`.
+- Similarly the entire derived class `Implementation` can be marked `final`. It has the same effect just for the entire class scope.
```
masterleinad, line 165:
```suggestion
- Similarly, the entire derived class `Implementation` can be marked `final`. It has the same effect just for the entire class scope.
```

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,60 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+- Tell the compiler to not look up any function name when calling `operator()` by using [qualified name lookup](https://en.cppreference.com/w/cpp/language/qualified_lookup). For this you tell the compiler which function you want by spelling out the class namespace e.g. `this->Implementation::operator() (i,v);`.
```
masterleinad, line 166:
```suggestion
- Tell the compiler not to look up any function name when calling `operator()` by using [qualified name lookup](https://en.cppreference.com/w/cpp/language/qualified_lookup). For this, you tell the compiler which function you want by spelling out the class namespace e.g. `this->Implementation::operator() (i,v);`.
```

docs/source/usecases/VirtualFunctions.md
JBludau: removed in [749fe2](https://github.com/kokkos/kokkos-core-wiki/pull/162/commits/749fe236c85d4ee7f9ed9417ccf160a081a17c40)

docs/source/usecases/VirtualFunctions.md
JBludau: I deleted the first sentence in [b0c6e4](https://github.com/kokkos/kokkos-core-wiki/pull/162/commits/b0c6e463f9c4bfbae4d48e1632851b4ec213cb89)

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,61 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+int main ()
+{
+    ... 
+    auto implementationPtr_h = std::make_shared<Implementation>();
+    implementationPtr_h->apply();
+}
+```
+### What is the problem?
+
+Inside the `parallel_for` the `operator()` is called which is marked `override`.
```
dalg24, line 150: Sentence reads weird

JBludau:
@dalg24 correctly mentioned, that the solution of making the function or class `final` is not guaranteeing to disable the V-Table lookup ... the standard does not define that it has to do this, so it is basically just another optimization that compilers use ... 
The qualified name lookup does not have this problem, it is strictly defined to circumvent the lookup (Virtual function call is suppressed if the function is selected using [qualified name lookup](https://en.cppreference.com/w/cpp/language/lookup) (that is, if the function's name appears to the right of the scope resolution operator ::).) from [cppreference](https://en.cppreference.com/w/cpp/language/virtual)

I will go ahead and mark the two 'solutions' with final as compiler dependent and reorder so the first is the qualified name lookup

docs/source/usecases/VirtualFunctions.md
JBludau: is [900707](https://github.com/kokkos/kokkos-core-wiki/pull/162/commits/9007071808750d512375291dbaedc433bbb1b9e8) better?

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,61 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+    ... 
+    auto implementationPtr_h = std::make_shared<Implementation>();
+    implementationPtr_h->apply();
+}
+```
+### What is the problem?
+
+Inside the `parallel_for` the `operator()` is called. As `Implementation` derives from the pure virtual class `Interface`, the 'operator()' is marked `override`.
+On ROCm 5.2 this results in a memory access violation.
+When executing the `this->operator()(i,v)` call, the runtime looks into the V-Table and dereferences a host pointer on the device.
```
PhilMiller, line 152: Slight mismatch
```suggestion
When executing the `this->operator()(i)` call, the runtime looks into the V-Table and dereferences a host pointer on the device.
```

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,61 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+
+### But if that is the case, why does it work with NVCC?
+
+Notice, that the `parallel_for` is called from a pointer of type `Implementation` and not a pointer of type `Interface` pointing to an `Implementation` object.
+Thus, no V-Table lookup for the `operator()` would be necessary as it can be deduced from the context of the call that it will be the `operator()` in the class `Interface`.
```
PhilMiller, line 157: Mistake, I think, and a more technically explicit wording:
```suggestion
Thus, no V-Table lookup for the `operator()` would be necessary as it can be deduced from the context of the call that it will be `Implementation::operator()`.
```

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,61 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+But here it comes down to how the compiler handles the lookup. NVCC understands that the call is coming from an `Implementation` object and thinks: "Oh, I see, that you are calling from an `Implementation` object, I know it will be the `operator()` in this class scope, I will do this for you".
+ROCm, on the other hand, sees your call and thinks “Oh, this is a call to a virtual method, I will look that up for you” - failing to read from the virtual function table, as it is containing host addresses.
```
PhilMiller, line 159: Technical accuracy and English wording:
```suggestion
ROCm, on the other hand, sees your call and thinks “Oh, this is a call to a virtual method, I will look that up for you” - failing to read from the virtual function table, as it resides in host space.
```
The memory access that fails is presumably to the v-table itself, not the dereference of the function pointers it contains.

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,61 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+
+### How to solve this?
+Strictly speaking, the observed behavior on NVCC is an optimization that uses the context information to avoid the V-Table lookup.
+If the compiler does not apply this optimization, you can help in different ways by providing additional information. 
+
+- Tell the compiler not to look up any function name in the V-Table when calling `operator()` by using [qualified name lookup](https://en.cppreference.com/w/cpp/language/qualified_lookup). For this, you tell the compiler which function you want by spelling out the class namespace e.g. `this->Implementation::operator() (i,v);`. This behavior is specified in the C++-Standard.
```
PhilMiller, line 165:
```suggestion
- Tell the compiler not to look up any function name in the V-Table when calling `operator()` by using [qualified name lookup](https://en.cppreference.com/w/cpp/language/qualified_lookup). For this, you tell the compiler which function you want by spelling out the class scope in which the function should be found e.g. `this->Implementation::operator() (i,v);`. This behavior is specified in the C++-Standard.
```

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,61 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+- Changing the `override` to `final` on the `operator()` in the `Implementation` class. This tells the compiler the `operator()` is not changing in derived objects. Many compilers do use this in optimization and deduce which function to call without the V-Table. Nevertheless, this might only work with certain compilers, as this effect of adding `final` is not specified in the C++-Standard.
```
PhilMiller, line 166:
```suggestion
- Changing the `override` to `final` on the `operator()` in the `Implementation` class. This tells the compiler the `operator()` is not changing in derived objects. Many compilers do use this in optimization and deduce which function to call without the V-Table. Nevertheless, this might only work with certain compilers, as this effect of adding `final` is not specified in the C++ Standard. 
```

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,61 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+
+### How to solve this?
+Strictly speaking, the observed behavior on NVCC is an optimization that uses the context information to avoid the V-Table lookup.
+If the compiler does not apply this optimization, you can help in different ways by providing additional information. 
+
+- Tell the compiler not to look up any function name in the V-Table when calling `operator()` by using [qualified name lookup](https://en.cppreference.com/w/cpp/language/qualified_lookup). For this, you tell the compiler which function you want by spelling out the class namespace e.g. `this->Implementation::operator() (i,v);`. This behavior is specified in the C++-Standard.
+- Changing the `override` to `final` on the `operator()` in the `Implementation` class. This tells the compiler the `operator()` is not changing in derived objects. Many compilers do use this in optimization and deduce which function to call without the V-Table. Nevertheless, this might only work with certain compilers, as this effect of adding `final` is not specified in the C++-Standard. 
+- Similarly, the entire derived class `Implementation` can be marked `final`. This is compiler dependent, too for the same reasons.
```
PhilMiller, line 167:
```suggestion
- Similarly, the entire derived class `Implementation` can be marked `final`. This is compiler dependent too, for the same reasons.
```

docs/source/usecases/VirtualFunctions.md
JBludau: goot catch, thanks

docs/source/usecases/VirtualFunctions.md
JBludau: even better catch

docs/source/usecases/VirtualFunctions.md
JBludau: thats what I meant but you are totally right

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,61 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+struct Implementation : public Interface
+{
+    KOKKOS_FUNCTION
+    void operator()(const size_t i) const override
+    { ... }
+
+    void apply(){
+        Kokkos::parallel_for("myLoop",10,
+            KOKKOS_CLASS_LAMBDA (const size_t i) { this->operator()(i); }
+        );}
```
dalg24, line 138:
```suggestion
        );
    }
```

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,61 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+};
+
+int main ()
+{
+    ... 
+    auto implementationPtr_h = std::make_shared<Implementation>();
+    implementationPtr_h->apply();
```
dalg24, line 145:
```suggestion
    auto implementationPtr = std::make_shared<Implementation>();
    implementationPtr->apply();
```

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,61 @@ deviceInstance->setAField(someHostValue); // set some field on the host
 
 This is the solution that the code teams we have talked to have said is the most productive way to solve the problem.
 
+## But what if I do not really need the V-Tables on the device side?
+
+The performance critical part of a code might not use pointers to base class and dynamic polymorphism to compute a result from data.
+Thus, the device might not need to have a working virtual function table.
+Consider the following example:
```
dalg24, line 118: Only beef I have what is being added is how it is introduced.  The current version reads somewhat confusing, I had to step back and remember what all this was about before reading on.

dalg24 (approved):

fnrizzi:
@PhilMiller i saw you reviewed this, can finalize approval please?

PhilMiller (approved):

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,60 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+But here it comes down to how the compiler handles the lookup. NVCC understands that the call is coming from an `Implementation` object and thinks: "Oh, I see, that you are calling from an `Implementation` object, I know it will be the `operator()` in this class scope, I will do this for you".
+ROCm, on the other hand, sees your call and thinks “Oh, this is a call to a virtual method, I will look that up for you” - failing to dereference the host function pointer in the host virtual function table.
+
+### How to solve this?
+Strictly speaking, the observed behavior on NVCC is an optimization that uses the context information to avoid the V-Table lookup.
+If the compiler does not apply this optimization, you can help in different ways by providing additional information. 
+
+- Tell the compiler not to look up any function name in the V-Table when calling `operator()` by using [qualified name lookup](https://en.cppreference.com/w/cpp/language/qualified_lookup). For this, you tell the compiler which function you want by spelling out the class scope in which the function should be found e.g. `this->Implementation::operator() (i);`. This behavior is specified in the C++ Standard.
```
masterleinad, line 164: Comment: But that doesn't mean that it would across backends. :slightly_smiling_face: E.g, `SYCL`, refuses to compile anything that says `virtual` but that's more a compiler/SYCL language problem.

docs/source/usecases/VirtualFunctions.md
```diff
@@ -111,6 +111,60 @@ deviceInstance->setAField(someHostValue); // set some field on the host
...
+        );
+    }
+};
+
+int main ()
+{
+    ... 
+    auto implementationPtr = std::make_shared<Implementation>();
+    implementationPtr->apply();
+}
```
masterleinad, line 145: I would prefer the example to actually compile.

docs/source/usecases/VirtualFunctions.md
JBludau: do you think we should address this? I feel like it is more of a shortcoming of `SYCL` but we could hint on it here ... even though we do not in the rest of the page (afaik)

masterleinad: No, I'm merely pointing out that support for `virtual` device functions is very brittle and there is not much `Kokkos` can do about it. We are at the mercy of the compiler/language.

docs/source/usecases/VirtualFunctions.md
JBludau: done in [6ba9f8](https://github.com/kokkos/kokkos-core-wiki/pull/162/commits/6ba9f8ee1ce63d6c7ec42203bbf02f3cd1d8acab)

docs/source/usecases/VirtualFunctions.md
masterleinad: The part that caught my intention is "This behavior is specified in the C++ standard" since `Kokkos` code needs to follow more the backend language than the C++ standard. Not all behavior specified by the C++ standard works the same in device code.

JBludau: I added a hint that it will not be fully portable in [ea04fc](https://github.com/kokkos/kokkos-core-wiki/pull/162/commits/ea04fccff66abf0ab72391f90b03ab739e13f699)

masterleinad (approved):

PR #178, opened 2022-10-03
single.md: responding to issue #5459

ajpowelsnl:
This PR represents a first draft of the `Kokkos::single` API documentation.  We placed it under `parallel_dispatch` because it is used within parallel regions.

### DIFF
--- a/docs/source/API/core/parallel-dispatch/single.md
+++ b/docs/source/API/core/parallel-dispatch/single.md
@@ -1,0 +1,84 @@
+# `fence()`
+
+Header File: `Kokkos_Core.hpp`
+
+Usage:
+
+```c++
+In a parallel dispatch statement, execute a lambda once per Thread or Team.
+- Kokkos::single` accepts two policies:
+  - Kokkos::PerTeam restricts execution of the lambda's body to once per team
+  - Kokkos::PerThread restricts execution of the lambda body to once per thread (i.e., to only one vector lane in a thread).
+
+```
+
+The `single` function can take two arguments:  an execution `Policy` and a `Lambda`.  `Kokkos::single` is only valid for the `Lambda` to capture variables by value (i.e., [=], and not [&]). That `Lambda` takes zero arguments or one argument by reference.  If `Lambda` takes one argument, the final value of that argument is broadcast to every executor, for example, every vector lane of the Thread, or every Thread (and vector lane) of the Team. 
+
+
+## Interface
+
+```c++
+Kokkos::single(Policy,Lambda);
+```
+
+### Parameters
+
+- `Policy`: Defines how the execution will be performed (in the context of a parallel dispatch statement)..
+
+### Requirements
+
+- `Kokkos::single(Policy,Lambda)` can be called inside an existing parallel region (i.e. inside the `operator()` of a functor or `Lambda`).
+
+## Semantics
+- Focuses execution in parallel regions to a single work item for the `Policy`.
+
+
+## Examples
+
+```
+using Kokkos::parallel_for;
+using Kokkos::parallel_reduce;
+using Kokkos::TeamThreadRange;
+using Kokkos::ThreadVectorRange;
+using Kokkos::PerThread;
+
+TeamPolicy<...> policy (...);
+typedef TeamPolicy<...>::member_type team_member;
+
+parallel_for (policy, KOKKOS_LAMBDA (const team_member& thread) {
+ // ...
+
+  parallel_for (TeamThreadRange (thread, 100),
+    KOKKOS_LAMBDA (const int& i) {
+      double sum = 0;
+      // Perform a vector reduction with a thread
+      parallel_reduce (ThreadVectorRange (thread, 100),
+        [=] (int i, double& lsum) {
+          // ...
+          lsum += ...;
+      }, sum);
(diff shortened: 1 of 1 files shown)


docs/source/API/core/parallel-dispatch/single.md
```diff
@@ -0,0 +1,84 @@
+# `fence()`
+
```
JBludau, line 2: why is there a `fence()` call here?

JBludau: If it was intended to be the title it should also contain a label so other pages can reference it.

docs/source/API/core/parallel-dispatch/single.md
```diff
@@ -0,0 +1,84 @@
...
+Usage:
+
+```c++
+In a parallel dispatch statement, execute a lambda once per Thread or Team.
+- Kokkos::single` accepts two policies:
+  - Kokkos::PerTeam restricts execution of the lambda's body to once per team
+  - Kokkos::PerThread restricts execution of the lambda body to once per thread (i.e., to only one vector lane in a thread).
+
+```
```
JBludau, line 13: I think the c++ block in this section should contain a 1-2 line call that conveys how it is used.

docs/source/API/core/parallel-dispatch/single.md
```diff
@@ -0,0 +1,84 @@
...
+
+The `single` function can take two arguments:  an execution `Policy` and a `Lambda`.  `Kokkos::single` is only valid for the `Lambda` to capture variables by value (i.e., [=], and not [&]). That `Lambda` takes zero arguments or one argument by reference.  If `Lambda` takes one argument, the final value of that argument is broadcast to every executor, for example, every vector lane of the Thread, or every Thread (and vector lane) of the Team.
```
JBludau, line 15:
```suggestion
The `single` function can take two arguments:  an execution `Policy` and a `Lambda`.  `Kokkos::single` is only valid if the `Lambda` captures variables by value (i.e., [=], and not [&]). The `Lambda` can take zero or one arguments by reference.  If it takes one argument, the final value of that argument is broadcast to every executor, for example, every vector lane of the Thread, or every Thread (and vector lane) of the Team. 
```

docs/source/API/core/parallel-dispatch/single.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Interface
+
+```c++
+Kokkos::single(Policy,Lambda);
+```
+
+### Parameters
+
+- `Policy`: Defines how the execution will be performed (in the context of a parallel dispatch statement)..
```
JBludau, line 26: I would either remove the list of one item or add the second argument and an explanation for it, even though it might seem obvious.

JBludau, line 26:
```suggestion
- `Policy`: Defines how the execution will be performed (in the context of a parallel dispatch statement).
```

docs/source/API/core/parallel-dispatch/single.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Requirements
+
+- `Kokkos::single(Policy,Lambda)` can be called inside an existing parallel region (i.e. inside the `operator()` of a functor or `Lambda`).
+
```
JBludau, line 31: The requirements are in the description. I think it would make sense to consolidate it somehow.

docs/source/API/core/parallel-dispatch/single.md
```diff
@@ -0,0 +1,84 @@
...
+## Semantics
+- Focuses execution in parallel regions to a single work item for the `Policy`.
```
JBludau, line 33: I would delete this section

docs/source/API/core/parallel-dispatch/single.md
```diff
@@ -0,0 +1,84 @@
...
+## Examples
+
+```
+using Kokkos::parallel_for;
+using Kokkos::parallel_reduce;
+using Kokkos::TeamThreadRange;
+using Kokkos::ThreadVectorRange;
+using Kokkos::PerThread;
+
+TeamPolicy<...> policy (...);
```
JBludau, line 45: could you make the example compilable?

docs/source/API/core/parallel-dispatch/single.md
```diff
@@ -0,0 +1,84 @@
...
+typedef TeamPolicy<...>::member_type team_member;
```
JBludau, line 46: I would not mix using and typedef

docs/source/API/core/parallel-dispatch/single.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+## Examples
+
+```
+using Kokkos::parallel_for;
+using Kokkos::parallel_reduce;
+using Kokkos::TeamThreadRange;
+using Kokkos::ThreadVectorRange;
+using Kokkos::PerThread;
```
JBludau, line 43: I think we should get rid of these here and use the namespace in the call as we do for `Single` in the example.

JBludau (changes requested):
I think it is more of in a draft state. Also needs to be added into the toc and alphabetical

ajpowelsnl:
Info not used

PR #188, opened 2022-10-19
algorithms: split non modifying seq ops

fnrizzi:
this PR splits into individual files the modifying seq ops. 
This is done for consistency with how all other algorithms are documented and also in preparation for documenting the team-level impl


fnrizzi:
@cz4rs i rebased this, so it is ready

cz4rs:
> @cz4rs i rebased this, so it is ready

Great, I will have a good look at it tomorrow.

cz4rs (approved):
Looks good to me.

PR #190, opened 2022-10-19
#189: Add `cppkokkos` extension to Sphinx

### LINKED ISSUE #189: Add `cppkokkos` extension to Sphinx
### TODO:
- add `cppkokkos` extension to Sphinx in order to allow further modifications

marcinwrobel1986:
### This PR:
- added cppkokkos extension
- modified Sphinx configuration

***IMPORTANT***: this is just the extension. Customizations will be applied in the next PR's.

#### Below an input:
```rst
``Kokkos::TestView``
====================

.. role:: cppkokkos(code)
   :language: cppkokkos

Header File: ``Kokkos_Core.hpp``

Class Interface
---------------

.. cppkokkos:class:: template <class DataType, class... Traits> TestView

  A potentially reference counted multi dimensional array with compile time layouts and memory space.
  Its semantics are similar to that of :cppkokkos:`std::shared_ptr`.

  .. rubric:: Public Member Variables

  .. cppkokkos:member:: static constexpr unsigned rank

    the rank of the view (i.e. the dimensionality).

  .. cppkokkos:type:: const_data_type

    The const version of ``DataType``, same as ``data_type`` if that is already const.

  .. cppkokkos:type:: uniform_runtime_type

    :cppkokkos:`uniform_type` but without compile time extents

  .. cppkokkos:function:: View( const ScratchSpace& space, const IntType& ... indices)

    *Requires:* :cppkokkos:`sizeof(IntType...)==rank_dynamic()` *and*  :cppkokkos:`array_layout::is_regular == true`.

    A constructor which acquires memory from a Scratch Memory handle.

    :param space: a scratch memory handle. Typically returned from :cppkokkos:`team_handles` in :cppkokkos:`TeamPolicy` kernels.
    :param indices: the runtime dimensions of the view.

```

#### Below an output:
Zrzut ekranu z 2022-10-20 09-58-11
Zrzut ekranu z 2022-10-20 09-58-19

closes #189


fnrizzi:
once we have the extension, can we mix directives? i mean can we do this:
```
.. cppkokkos:type:: const_data_type

    bla blas

.. cpp:type:: const_data_type

    bla bla
```

marcinwrobel1986:
> @marcinwrobel1986 once we have the extension, can we mix directives? i mean can we do this:
> 
> ```
> .. cppkokkos:type:: const_data_type
> 
>     bla blas
> 
> .. cpp:type:: const_data_type
> 
>     bla bla
> ```

I am not sure I need to check that, because at the top the language domain is defined. I will be back with an answer.

marcinwrobel1986:
@fnrizzi `cpp` domain will works as well.
Below the input:
```rst
``Kokkos::TestView``
====================

.. role:: cppkokkos(code)
   :language: cppkokkos

.. role:: cpp(code)
   :language: cpp

Header File: ``Kokkos_Core.hpp``

Class Interface
---------------

.. cppkokkos:class:: template <class DataType, class... Traits> TestView

  A potentially reference counted multi dimensional array with compile time layouts and memory space.
  Its semantics are similar to that of :cppkokkos:`std::shared_ptr`.

  .. rubric:: Public Member Variables

  .. cppkokkos:member:: static constexpr unsigned rank

    the rank of the view (i.e. the dimensionality).

  .. cppkokkos:type:: const_data_type

    The const version of ``DataType``, same as ``data_type`` if that is already const.

  .. cppkokkos:type:: uniform_runtime_type

    :cppkokkos:`uniform_type` but without compile time extents

  .. cppkokkos:function:: View( const ScratchSpace& space, const IntType& ... indices)

    *Requires:* :cppkokkos:`sizeof(IntType...)==rank_dynamic()` *and*  :cppkokkos:`array_layout::is_regular == true`.

    A constructor which acquires memory from a Scratch Memory handle.

    :param space: a scratch memory handle. Typically returned from :cppkokkos:`team_handles` in :cppkokkos:`TeamPolicy` kernels.
    :param indices: the runtime dimensions of the view.

.. cpp:class:: template <class DataType, class... Traits> TestView

  A potentially reference counted multi dimensional array with compile time layouts and memory space.
  Its semantics are similar to that of :cpp:`std::shared_ptr`.

  .. rubric:: Public Member Variables

  .. cpp:member:: static constexpr unsigned rank

    the rank of the view (i.e. the dimensionality).

  .. cpp:type:: const_data_type

    The const version of ``DataType``, same as ``data_type`` if that is already const.

  .. cpp:type:: uniform_runtime_type

    :cpp:`uniform_type` but without compile time extents

  .. cpp:function:: View( const ScratchSpace& space, const IntType& ... indices)

    *Requires:* :cpp:`sizeof(IntType...)==rank_dynamic()` *and*  :cpp:`array_layout::is_regular == true`.

    A constructor which acquires memory from a Scratch Memory handle.

    :param space: a scratch memory handle. Typically returned from :cpp:`team_handles` in :cpp:`TeamPolicy` kernels.
    :param indices: the runtime dimensions of the view.
```

Below the output:
Zrzut ekranu z 2022-10-20 12-18-20
Zrzut ekranu z 2022-10-20 12-18-37

marcinwrobel1986:
> I think we will need enforce using `cppkokkos` always otherwise IMO having those lines at the top of each file become too much of a burden. Also, always using cppkokkos should be the way to go

I agree! `cppkokkos` has everything that `cpp` domain has. I think the same, that using both could be confusing.

fnrizzi (approved):

fnrizzi:
@crtrott @dalg24 @masterleinad @nmm0 @JBludau since this is just the extension, if you agree on the name "cppkokkos", we should merge it. Once this is merged, the regular cpp domain will still work so we can incrementally add new customizations (#191 #192 ) without breaking anything

JBludau (commented):
We still need to think about how to document the changes we did to the cpp extension. @marcinwrobel1986 did you add extensive comments to the cppkokkos.py on all places you changed something? And do you have a searchable marker so all the changes can be found easily?

fnrizzi:
> We still need to think about how to document the changes we did to the cpp extension. @marcinwrobel1986 did you add extensive comments to the cppkokkos.py on all places you changed something? And do you have a searchable marker so all the changes can be found easily?

yes we need to do that, but that can wait. We need to first *decide and approve* a basic set customizations (#191 #192) and then we can make a simple reference to start with so that each developer can use when writing documentation

JBludau:
I guess I am fine with it if we create an issue to track that we still need to add comments to the places in the extension that @marcinwrobel1986 customized

marcinwrobel1986:
> 

Hello @JBludau I just want to mention once more, that `cppkokkos` domain is a separate thing. It is totally isolated from Sphinx itself and changes there has nothing to do with the Sphinx code. In `cppkokkos` domain definitions of how Sphinx should interpretate this domain exists. Knowing that we could OFC mark where the changes to base (which was `cpp` domain) were made.

JBludau:
> Knowing that we could OFC mark where the changes to base (which was `cpp` domain) were made.

That is all I would need to be happy. But I am fine with doing it later if you prefer (meaning after you changed more stuff to integrate the other changes we are hoping for). I just don't want it to be forgotten and someone in the far future would have to extensively diff the files to the original `cpp` domain to make sense of your changes.

dalg24:
The changes between cppkokkos and the cpp needs to be made obvious.  Add the original file as is in a first commit and apply changes in a 2nd commit.  As suggested by Jakob consider adding greppable markers to find the location of the changes (optional)

dalg24 (changes requested):
Changes from origin must be made obvious

nmm0 (commented):
Is there a way to import the cpp domain and re-use as much of it as possible aside from our custom directives? That way we can avoid having to update 8000 lines of code if the base domain changes

marcinwrobel1986:
> Is there a way to import the cpp domain and re-use as much of it as possible aside from our custom directives? 

`cpp` domain could be used in parallel with `cppkokkos` but as Francesco wrote previously this will bring only the confusion, and I agree with that. One can not registered two domains with the same name. We need mostly same behavior as in `cpp` domain, so everything in `cpp` domain has to be implemented in `cppkokkos`

> That way we can avoid having to update 8000 lines of code if the base domain changes

Once again guys our `cppkokkos` domain is separated from `cpp` domain. Any changes in `cpp` we do not have to update our `cppkokkos` at all. When we move to Python 3.11 which was released on 24th Oct, then we could forgot about updates till 24 Oct 2027. The only reason to upgrade would be introducing any new keywords, operators and so on to C++ language.

FYI the other way would be to create `cppkokkos` from scratch, which would be reinventing the wheel again and in case of introducing any new keywords, operators and so on to C++ language that would have to be added anyway.

The thing is that `cpp` domain is very specific, it has ca. 8k lines, where `c` domain has 4k, `python` domain has 1.5k lines and `javascript` has 450 lines

marcinwrobel1986:
Hello @fnrizzi @dalg24 @nmm0 @JBludau 

1. I reverted the initial commit.
2. I pushed a commit with unmodified version of `cpp` domain
3. I changed `cpp` domain to `cppkokkos` and pushed committed changes
4. I marked changes with:
  - `@@@@@_` as beginning of the changes made
  - `@@@@@!` as the end of the changes made
5. I pushed a commit with added markers

crtrott (approved):

PR #204, opened 2022-11-08
#120: API core view LayoutRight transition

marcinwrobel1986:
* API/core/view/LayoutRight transition from .md to .rst

before (.md) | after (.rst)
:----------------: | :-------------:
image | image

PR #216, opened 2022-12-01
#120: API-core-initialize_finalize-finalize from md to rst

pierrepebay:
### **THIS PR:**
* API/core/initialize_finalize/finalize transition from .md to .rst

before (.md) | after (.rst)
before | after


dalg24:
To make sure I understand: The only observable difference is the `finalize` header but that is presumably intentional since the header is in code font in the .rst file.

fnrizzi:
@dalg24 correct, because this file is simple. The major differences will be mostly for files containing cpp classes that will need to be converted to the Sphinx cpp class domain/syntax. But converting everything to rst allows us to setup everything so that if later we need to change/add things, it will be already ready to use Sphinx

fnrizzi:
And if for example we want to customize the rendering of a cpp function via 
```
cppkokkos::function
```
We have everything ready to support that

fnrizzi:
I just did above , we overlapped

dalg24:
OK I commented before I saw Francesco's reply.  I assume the rational is "so that we can use the customized rendering everywhere".

docs/source/API/core/initialize_finalize/finalize.rst
```diff
@@ -0,0 +1,56 @@
...
+Once this function is called, no Kokkos API functions (not even
+`Kokkos::initialize <initialize.html#kokkosinitialize>`_) may be called, except for
+``Kokkos::is_initialized`` or ``Kokkos::is_finalized``.
+The user must ensure that all Kokkos objects (e.g. ``Kokkos::View``) are detroyed
+before ``Kokkos::finalize`` gets called.
+
+Programs are ill-formed if they do not call this function after calling `Kokkos::initialize <initialize.html#kokkosinitialize>`_.
+
+Interface
+============
```
dalg24, line 27:
```suggestion
Interface
=========
```

fnrizzi:
And the other reason is that we said this way we have a single consistent way of doing things

dalg24 (approved):

fnrizzi:
@pierrepebay one small comment: for consistency, please use the same symbol to identify sections, subsections ,etc as in the other PRs see this for example: https://github.com/kokkos/kokkos-core-wiki/pull/204/files

fnrizzi (approved):

PR #218, opened 2022-12-08
Add nested sort to Sort wiki page

brian-kelley:
Document new functions in Kokkos_NestedSort.hpp: sort_team, sort_thread_by_key, etc. with full working example.

Adding the people who reviewed the original code PR

Can see the rendered markdown in my branch here: https://github.com/brian-kelley/kokkos-core-wiki/blob/AddNestedSort/docs/source/API/algorithms/Sort.md#sorting-with-nested-policies-team--and-thread-level


docs/source/API/algorithms/Sort.md
```diff
@@ -29,3 +29,152 @@ template<class KeyViewType>  struct BinOp3D  { }
...
+```
+
+``sort_team`` and ``sort_by_key_team`` internally use the entire team, so they may be called inside the top level of ``TeamPolicy`` lambdas and functors. ``sort_thread`` and ``sort_by_key_thread`` use the vector lanes of a thread, so they may be called either within ``TeamPolicy`` or ``TeamThreadRange`` loops.
+
+The ``sort_by_key`` functions sort ``keyView``, while simultaneously applying the same permutation to the elements of ``valueView``. It is equivalent to sorting ``(key[i], value[i])`` tuples according by key. An example of where this is commonly used is to sort the entries and values within each row of a CRS (compressed row sparse) matrix. These functions require that ``keyView.extent(0) == valueView.extent(0)``.
+
+Versions taking a ``Comparator`` object will use it to order the keys. ``Comparator::operator()`` should be a const member function that accept two keys ``a`` and ``b``, and returns a bool that is true if and only if ``a`` goes before ``b`` in the sorted list. For versions not taking a ``Comparator`` object, keys are sorted into ascending order (according to ``operator<``). For example, this comparator will sort a view of ``int`` in _descending_ order:
```
masterleinad, line 79:
```suggestion
Versions taking a ``Comparator`` object will use it to order the keys. ``Comparator::operator()`` should be a const member function that accepts two keys ``a`` and ``b``, and returns a bool that is true if and only if ``a`` goes before ``b`` in the sorted list. For versions not taking a ``Comparator`` object, keys are sorted into ascending order (according to ``operator<``). For example, this comparator will sort a view of ``int`` in _descending_ order:
```

masterleinad (approved):

PR #221, opened 2022-12-09
Minor edits to nested sort

brian-kelley:
Fix a few places with unclear/imprecise wording


fnrizzi (approved):

docs/source/API/algorithms/Sort.md
```diff
@@ -166,7 +166,7 @@ A, sorted by row:
 4 19 20 29 42 56 60 63 68 90 
 1 16 16 17 33 39 60 64 78 94 
 
-A, now sorted by col:
+A, now with each column sorted:
```
fnrizzi, line 169: really picky comment, but i would remove the "now" since for the comment for the row above does not have it

PR #224, opened 2022-12-13
add missing aliases to team_handle_concept page

fnrizzi:
the original concept page was missing nested aliases

### DIFF
--- a/docs/source/API/core/policies/TeamHandleConcept.md
+++ b/docs/source/API/core/policies/TeamHandleConcept.md
@@ -14,6 +14,10 @@ and are thus subject to change.
 ```c++
 class TeamHandleConcept {
     public:
+    // Nested aliases
+    using execution_space      = /*see details in the description below*/;
+    using scratch_memory_space = /*see details in the description below*/;
+
     // Constructor, Destructor, Assignment
     TeamHandleConcept( TeamHandleConcept && ) = default ;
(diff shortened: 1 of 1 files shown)


docs/source/API/core/policies/TeamHandleConcept.md
```diff
@@ -59,6 +63,12 @@ class TeamHandleConcept {
 
 ## Public Class Members
 
+### Nested aliases
+
+  * `execution_space`: specifies the [execution space](https://kokkos.github.io/kokkos-core-wiki/API/core/execution_spaces.html) associated to the team
```
dalg24, line 68: Why is it a full URL rather than some sort of relative path?

fnrizzi: Because I will have to convert the page to rst anyway, so I did not want to put a relative path that works for me only and then change it again. I will make it relative in rst directly

dalg24: I am a bit surprised that it is not possible regardless of whether we use rst or md but fine.

dalg24 (approved):

docs/source/API/core/policies/TeamHandleConcept.md
fnrizzi: it is possible, just that it would have two different syntaxes

PR #266, opened 2023-01-16
#120: API-core-numerics-mathematical_functions from md to rst

antoinemeyer5:
### THIS PR:
* API/core/numerics/mathematical-functions transition from .md to .rst

- [x] uses cppkokkos
- [x] rebased on latest main

before (.md) | after (.rst)
image | image
image | image
image | image


docs/source/API/core/numerics/mathematical-functions.rst
```diff
@@ -0,0 +1,215 @@
...
+
+The function is overloaded for any argument of arithmetic type. Additional functions with ``f`` and ``l`` suffixes that work on ``float`` and ``long double`` respectively are also available.  Please note, that ``long double`` overloads are not available on the device.
+
+See below the list of common mathematical functions supported. We refer the reader to cppreference.com for the synopsis of each individual function.
+
+------------
+
+:strike:`func` denotes functions that are currently not provided by Kokkos
+
+``func*`` see notes below
```
ajpowelsnl, line 64: Would it be possible to simply separate the supported vs. not supported functions with a heading?  Or even just list the supported functions, making no statement about what is not supported?  That would solve the strikethrough issue.  If Sphinx doesn't like `X` as a variable name, simply call it VariableX, or something like that.  Does this solution make sense in the context of the issues you're having?

ajpowelsnl (changes requested):
Hi Antoine - I've suggested a solution.  Please let me know if it's not workable for the problems you're seeing.

docs/source/API/core/numerics/mathematical-functions.rst
antoinemeyer5: Hello @ajpowelsnl 

Solution 1 to separate and put a title is possible.
Solution 2 to remove the unsupported functions is possible.

It's a matter of style choice, I let you choose what you prefer.

Choice 1.1
```
# Basic operations
## Provided
abs, fabs, [and more]
## Not provided
remquo
```

Choice 1.2: Header # Not provided only here if applicable
```
# Basic operations
abs, fabs, [and more]
# Not provided
remquo
```

Choice 2 
```
# Basic operations
abs, fabs, [and more]

# Exponential functions
exp, exp2, [and more]
```

antoinemeyer5: @ajpowelsnl 
The problem is that Sphinx does not like mixing notations. A variable should either be bold, italic or strikethrough, but Sphinx doesn't like bold + italic or italic + strikethrough variables.

(My examples with x are bad, I just wanted to show the notation \`\` or ** or strikethough)

antoinemeyer5: Currently:
file .md | file .rst
image | image

I'm waiting for your choice @fnrizzi & @ajpowelsnl  of style to finish!

ajpowelsnl: @antoinemeyer5 and @fnrizzi - I prefer showing what we support.  Showing what we do not support (even though it's part of the Standard) is a bit odd.  This approach takes care of the strikethrough problem, doesn't it?

PhilMiller: If we've implemented a subset of the standard, it's much easier on downstream developers looking for a function to see an explicit "not implemented yet", maybe with a link to where/how to request it, than to just not find anything, and have to ask/discover elsewhere what we say about missing functions.

ajpowelsnl: @PhilMiller - I can live with that -- good way to tee up work, too.

antoinemeyer5: My proposal:
live .md | rework .rst
image |  image

I'm waiting for your opinion @fnrizzi & @ajpowelsnl & @PhilMiller to finish

fnrizzi (approved):

ajpowelsnl (approved):

fnrizzi (approved):

PR #273, opened 2023-01-24
Fixup deducing `Kokkos::Impl::ALL_t` in `Kokkos::Subview` doc

dalg24:
Code snippets do not compile.  Reported on Slack.

Opening just a base for discussion as we could probably update the definition of the subview and make it work instead.

### DIFF
--- a/docs/source/API/core/view/subview.md
+++ b/docs/source/API/core/view/subview.md
@@ -44,7 +44,7 @@ IMPL_DETAIL subview(const ViewType& v, Args ... args);
     * `std::pair<iType,iType>` with `std::is_integral<iType>::value` being true.
     * `Kokkos::pair<iType,iType>` with `std::is_integral<iType>::value` being true.
     * `iType` with `std::is_integral<iType>::value` being true.
-    * `decltype(`[`Kokkos::ALL`](KokkosALL)`)`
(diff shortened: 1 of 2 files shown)


dalg24:
The alternative is
```diff
diff --git a/core/src/impl/Kokkos_ViewMapping.hpp b/core/src/impl/Kokkos_ViewMapping.hpp
index 3ab8237cd..2956d5727 100644
--- a/core/src/impl/Kokkos_ViewMapping.hpp
+++ b/core/src/impl/Kokkos_ViewMapping.hpp
@@ -3805,7 +3805,9 @@ struct SubViewDataTypeImpl<
           Kokkos::Experimental::Extents<Exts...>, Args...> {};

 template <class ValueType, class Exts, class... Args>
-struct SubViewDataType : SubViewDataTypeImpl<void, ValueType, Exts, Args...> {};
+struct SubViewDataType
+    : SubViewDataTypeImpl<void, ValueType, std::remove_const_t<Exts>,
+                          std::remove_const_t<Args>...> {};

 //----------------------------------------------------------------------------

masterleinad:
My preference would be to make the example as it is work.

crtrott:
Fix the documentation and then make ALL_t public ...

crtrott (approved):

PR #290, opened 2023-02-01
Fixup `Kokkos_ENABLE_PROFILING` option was removed in 3.2 release

dalg24:


docs/source/keywords.rst
```diff
@@ -143,10 +143,6 @@ Enable Options
       * Perform extra large memory tests
       * ``OFF``
 
-    * * ``Kokkos_ENABLE_PROFILING``
-      * Create bindings for profiling tools
-      * ``ON``
-
     * * ``Kokkos_ENABLE_PROFILING_LOAD_PRINT``
```
fnrizzi, line 146: this load_print one is not dependent on the one you removed?

dalg24: No.  I suppose it is dependent on whether LIBDL is enabled or not.

fnrizzi: so I can turn on profiling_load_print without the other , right? no issues? I guess not since it was removed since 3.2 but just asking since I do not know

dalg24: I am not sure I understand the question.
`Kokkos_ENABLE_PROFILING` is not used.  `Kokkos_ENABLE_LIBDL` must be `ON` in order to use the kokkos tools, meaning `Kokkos_ENABLE_PROFILING_LOAD_PRINT` has no effect if `Kokkos_ENABLE_LIBDL` is `OFF`.  I haven't looked whether we warn about this.

fnrizzi (approved):

PR #299, opened 2023-02-20
fix `kokkosinlinefunction` to support attributes

fnrizzi:
The current impl of `kokkosinlinefunction` was only partially working. 
This PR fixes it  so that we can prepend static, etc 

```
.. role:: cppkokkos(code)
    :language: cppkokkos

.. cppkokkos:kokkosinlinefunction:: double foo1(int const a);

.. cppkokkos:kokkosinlinefunction:: template<class T> double foo1(int const a);

.. cppkokkos:kokkosinlinefunction:: static double foo2(int const a);

.. cppkokkos:kokkosinlinefunction:: explicit constexpr double foo(int a = 0);

```

Renders as: 

image

### DIFF
--- a/docs/source/_ext/cppkokkos.py
+++ b/docs/source/_ext/cppkokkos.py
@@ -3707,7 +3707,7 @@ class ASTDeclaration(ASTBase):
 
     @property
     def function_params(self) -> List[ASTFunctionParameter]:
-        if self.objectType != 'function':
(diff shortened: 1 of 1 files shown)


ajpowelsnl (approved):
Looks good

nmm0 (approved):
LGTM!

PR #300, opened 2023-02-21
Add documentation for `<Kokkos_BitManipulation.hpp>` header

dalg24:
Corresponds to kokkos/kokkos#4577

### DIFF
--- a/docs/source/API/core/Numerics.rst
+++ b/docs/source/API/core/Numerics.rst
@@ -24,3 +24,11 @@ library mathematical functions, such as ``fabs``, ``sqrt``, and ``sin``.
 
 The header ``<Kokkos_NumericTraits.hpp>`` implements a new facility that is being added to the C++23 standard library and
 is intended as a replacement for ``std::numeric_limits``.
+
+.. toctree::
+   :maxdepth: 1
+
+   numerics/bit-manipulation.md
+
+The header ``<Kokkos_BitManipulation.hpp>`` is a backport of the C++20 standard library header ``<bit>`` and
+provides several function templates to access, manipulate, and process individual bits and bit sequences.
--- a/docs/source/API/core/numerics/bit-manipulation.rst
+++ b/docs/source/API/core/numerics/bit-manipulation.rst
@@ -1,0 +1,102 @@
+Bit manipulation
+================
+
+.. role::cpp(code)
+    :language: cpp
+
+.. role:: strike
+    :class: strike
+
+.. _KokkosBitManipulation: https://github.com/kokkos/kokkos/blob/4.1.00/core/src/Kokkos_BitManipulation.hpp
+
+.. |KokkosBitManipulation| replace:: ``<Kokkos_BitManipulation.hpp>``
+
+.. _StandarLibraryHeaderBit: https://en.cppreference.com/w/cpp/header/bit
+
+.. |StandarLibraryHeaderBit| replace:: ``<bit>``
+
+Defined in header |KokkosBitManipulation|_ which is included from ``<Kokkos_Core.hpp>``
+
+Provides function templates from the standard library header |StandarLibraryHeaderBit|_ (since C++20).
+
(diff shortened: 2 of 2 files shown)


docs/source/API/core/numerics/bit-manipulation.rst
```diff
@@ -0,0 +1,102 @@
...
+
+.. |StandarLibraryHeaderBit| replace:: ``<bit>``
+
+Defined in header |KokkosBitManipulation|_ which is included from ``<Kokkos_Core.hpp>``
+
+Provides function templates from the standard library header |StandarLibraryHeaderBit|_ (since C++20).
+
+Bit manipulation function templates are defined in the ``Kokkos::`` namespace since Kokkos 4.1
+
+.. _bit_cast: https://en.cppreference.com/w/cpp/numeric/bit_cast
```
masterleinad, line 24: This doesn't appear as a link.

docs/source/API/core/numerics/bit-manipulation.rst
```diff
@@ -0,0 +1,102 @@
...
+
+.. |bit_cast| replace:: ``bit_cast``
+
+.. _byteswap: https://en.cppreference.com/w/cpp/numeric/byteswap
```
masterleinad, line 28: This doesn't appear as a link. This is C++23.

docs/source/API/core/numerics/bit-manipulation.rst
```diff
@@ -0,0 +1,102 @@
...
+
+.. _bit_floor: https://en.cppreference.com/w/cpp/numeric/bit_floor
+
+.. |bit_floor| replace:: ``bit_floor``
+
+.. _bit_width: https://en.cppreference.com/w/cpp/numeric/bit_width
+
+.. |bit_width| replace:: ``bit_width``
+
+.. _rotl: https://en.cppreference.com/w/cpp/numeric/rotl
```
masterleinad, line 48: This doesn't appear as a link.

docs/source/API/core/numerics/bit-manipulation.rst
```diff
@@ -0,0 +1,102 @@
...
+
+.. |rotl| replace:: ``rotl``
+
+.. _rotr: https://en.cppreference.com/w/cpp/numeric/rotr
```
masterleinad, line 52: This doesn't appear as a link.

docs/source/API/core/numerics/bit-manipulation.rst
dalg24: I know.  Don't think we can render them crossed out and had them in anticipation of adding later.  Want me to remove?

masterleinad: No, I think it's good to already have the links but maybe can find another way to mark unimplemented functions.

dalg24: > maybe can find another way to mark unimplemented functions

any suggestion?

docs/source/API/core/Numerics.rst
```diff
@@ -24,3 +24,11 @@ library mathematical functions, such as ``fabs``, ``sqrt``, and ``sin``.
 
 The header ``<Kokkos_NumericTraits.hpp>`` implements a new facility that is being added to the C++23 standard library and
 is intended as a replacement for ``std::numeric_limits``.
+
+.. toctree::
+   :maxdepth: 1
+
+   numerics/bit-manipulation.md
```
ldh4, line 31: Shouldn't this be `numerics/bit-manipulation.rst`?

dalg24: That is a good point.  I copy/pasted from the same file without paying attention.  Want me to fix here or ok to deal with it as a follow up?

ldh4: Ah looks like the links in `Numerics.rst` haven't been replaced along with the md to rst file conversions then. I think it's fine to fix those as a follow up.

dalg24:
Ping

janciesko (approved):

PR #308, opened 2023-02-21
#120: API-core-stl_compat-pair from md to rst

antoinemeyer5:
### THIS PR:
* API/core/stl_compat/pair transition from .md to .rst

- [x] uses cppkokkos
- [x] rebased on latest main
- [x] screenshots

before (.md) | after (.rst)
image | image
image | image


fnrizzi (changes requested):
@antoinemeyer5 
1) please remote the synopsis 
2) please document the class properly using cppkokkos, right now it is a mix of blocks and things

antoinemeyer5:
I always get error messages.

Here are 2 examples that work and the result we'd like to have that doesn't:

image

The error comes from the fact that the return type of the function contains `<` and `>`.

What should I do now @fnrizzi?

PR #325, opened 2023-03-02
Add NVIDIA Ada GPU architecture

dalg24:
Added in kokkos/kokkos#5867


ajpowelsnl (approved):
Looks good.  The current version (https://kokkos.github.io/kokkos-core-wiki/API/core/Macros.html) does not have the table of architectures (I see the "Architectures" heading, but no entries).

PR #326, opened 2023-03-02
Remove macro and option `Kokkos_ENABLE_HPX_ASYNC_DISPATCH`

dalg24:
Became implementation detail in kokkos/kokkos#5628


dalg24:
Considering it is an experimental backend I thought simple removal keeps things simple.
@msimberg any opinion on this?

msimberg (approved):
I'm fine with removing (documentation about) it.

All users of the HPX backend that I know of are already aware that it's been changed to `IMPL`, and it probably should've been `IMPL` from the beginning.

Edit: No biggie, but maybe the title of the PR should include "from documentation"? The option still exists, just as `IMPL`.

PR #330, opened 2023-03-03
Fix typo

PhilMiller:


dalg24 (approved):
Looks good but considering that the file hasn't been migrated to rst I will let Francesco decide how to handle this

fnrizzi (approved):

PR #331, opened 2023-03-03
Update Intel architectures

masterleinad:


docs/source/keywords.rst
```diff
@@ -306,8 +306,32 @@ Architecture Keywords
       * Optimize for HSW architecture
       * ``OFF``
 
-    * * ``Kokkos_ARCH_INTEL_GEN``
-      * Optimize for Intel GPUs Gen9+
+    * * ``KOKKOS_ARCH_INTEL_GEN``
+      * Optimize for Intel GPUs, JIT
```
dalg24, line 310: Do we explain somewhere what JIT means and how it differs from other options?

docs/source/keywords.rst
```diff
@@ -378,6 +402,10 @@ Architecture Keywords
       * Optimize for SNB architecture
       * ``OFF``
 
+    * * ``Kokkos_ARCH_SPR``
+      * Optimize for Sapphire Rapids architecture
```
dalg24, line 406: Say Intel

dalg24: Nevermind I see we don't say anything for other Intel CPU architecture.  We can improve later.

docs/source/keywords.rst
masterleinad: No, as far as I know, this is the only occurrence of something JIT related in `Kokkos`. Do you think it's sufficient to just spell it out here?

dalg24: You will need to write a paragraph that explains the alternatives ahead-of-time VS just-in-time and what these entail.
Adding as a follow-up is fine if you prefer

masterleinad: Fair enough. I gave it a shot. 🙂

dalg24 (approved):

Rombur (approved):

docs/source/API/core/Macros.rst
```diff
@@ -254,3 +256,8 @@ Architectures
 | ``KOKKOS_ARCH_VEGA90A``           | Optimize for AMD GPU MI200 series GFX90A (GPU)                                    |
 +-----------------------------------+-----------------------------------------------------------------------------------+
 
+*) ``KOKKOS_ARCH_INTEL_GEN`` enables Just-In-Time compilation for Intel GPUs whereas all the other flags for Intel compilers
+request Ahead-Of-Time compilation. Just-In-Time compilation means that the compiler is invoked again when the binaries created
+are actually executed and only at that point the architecture to compile for is determined. On the other hand, Ahead-Of-Time
+compilation describes the standard model where the compiler is only invoked once to create the binary and the architecture to
+compile for is determined before the program is run.
```
dalg24, line 263: I just realized you commented by the macros.  It is actually more useful by the CMake options

PR #333, opened 2023-03-03
Update compiler requirements for 4.0

### LINKED ISSUE #310: Update 4.0 compiler requirements
https://github.com/kokkos/kokkos-core-wiki/blob/main/docs/source/requirements.rst is out of date for 4.0

We should try and do this when we version 3.7/4.0 if we can. Otherwise we really need to make sure this is up to date

crtrott:
Fix #310

### DIFF
--- a/docs/source/requirements.rst
+++ b/docs/source/requirements.rst
@@ -4,9 +4,9 @@ Requirements
 Compiler Versions
 =================
 
-Generally Kokkos should work with all compiler versions newer than the minimum.
+Generally, Kokkos should work with all compiler versions newer than the minimum.
 We also recommend using updated compilers where possible to benefit from improvements.
-However,in complex code, we have to work around compiler bugs. So compiler versions we don't test may have issues we are unaware of.
+However, in complex code, we have to work around compiler bugs. So compiler versions we don't test may have issues we are unaware of.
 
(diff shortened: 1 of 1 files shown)


docs/source/requirements.rst
```diff
@@ -5,8 +5,72 @@ Compiler Versions
 =================
 
 Generally Kokkos should work with all compiler versions newer than the minimum.
```
masterleinad, line 7:
```suggestion
Generally, Kokkos should work with all compiler versions newer than the minimum.
```

docs/source/requirements.rst
```diff
@@ -5,8 +5,72 @@ Compiler Versions
...
-However,in complex code, we have to work around compiler bugs. So compiler versions we don't test may have issues
-we are unaware of.
+We also recommend using updated compilers where possible to benefit from improvements.
+However,in complex code, we have to work around compiler bugs. So compiler versions we don't test may have issues we are unaware of.
```
masterleinad, line 9:
```suggestion
However, in complex code, we have to work around compiler bugs. So compiler versions we don't test may have issues we are unaware of.
```

dalg24, line 9: Would prefer a space after the comma and breaking long lines

dalg24 (approved):

PR #338, opened 2023-03-05
#120: API-algorithms-std_algorithms-StdModSeq from rst to new rst

antoinemeyer5:
### THIS PR:
* API/algorithms/std-algorithms/StdModSeq transition from .rst to new .rst

before (.md) | after (.rst)
image | image


fnrizzi:
@antoinemeyer5 why did you change this file? this was already an rst and it is only including files. There was no reason to change this :)

PR #346, opened 2023-03-10
fix error in View constructor

fnrizzi:
BEFORE 

image

AFTER:

image


dalg24:
So I was thinking it is stupid to pass the pointer by const reference both in that constructor and in the one above, and I thought whatever I will update when I fix the code but turns out the implementation takes the pointer by value.
Can you please update the doc to reflect that?

fnrizzi:
image

dalg24 (approved):

Issue #272, opened 2023-01-23
Documentation versioning TODOs


nmm0:
Some things that will need to be done when we split off Kokkos 3.7 from 4.0 documentation. Please feel free to add or create a new issue:

- [ ] [4.0 changes here need to be versioned](https://github.com/kokkos/kokkos-core-wiki/blob/main/docs/source/API/core/numerics/numeric-traits.md)

ajpowelsnl:
@nmm0 , @fnrizzi , @dalg24 -- Is this still something we want to pursue at this time?  If so, let's create a sensible plan to do it.  If not, can we close this issue and define the condition /  point at which it would make sense to start versioning?

PR #352, opened 2023-03-30
[Draft] initial versioning using sphinx-multiversioning

nmm0:
https://github.com/kokkos/kokkos-core-wiki/issues/272 needs to be updated/completed before merging this

Right now `test` is whitelisted as a valid branch name for versioning, you can create a local `test` branch to try it out. When we merge, we should probably remove that.

There are some changes that will need to be done with the CI to get this to work properly. We also need to have a branch (and ensure that it is pulled in the CI) for 3.7 and 4.0.

Right now the multiversioning uses "main" for the main branch, but "latest" is more conventional. I can't figure out a good way of getting it to use "latest" without renaming the branch.

### More info

sphinx-multiversioning works by reading all tags and branches in the *local* git repository. I've currently whitelisted the branches "main" and "test" (just for testing purposes) and any tag that is of the format "x.x.x.x".

It will generate a subfolder with the git ref name under the generated_docs folder

I had to tweak the `edit_button_handler.py` script to work with multiple versions, but fortunately sphinx-multiversioning outputs a json file that contains all the generated versions, so I just read that.

### DIFF
--- a/build_requirements.txt
+++ b/build_requirements.txt
@@ -3,4 +3,4 @@ furo==2022.6.4.1
 myst-parser==0.18.0
 sphinx-copybutton==0.5.0
 m2r2==0.3.2
-sphinx-multiversion=0.2.4
+sphinx-multiversion==0.2.4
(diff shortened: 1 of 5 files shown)


ajpowelsnl:
We anticipate versioning ca. Kokkos-5.  Until that time, our focus is to complete API Guide updates , filling any gaps.

PR #355, opened 2023-04-05
Document default value `{OFF -> ON}` change in 4.0 for `Kokkos_ENABLE_CUDA_LAMBDA` option

dalg24:


masterleinad (approved):

docs/source/keywords.rst
```diff
@@ -93,7 +93,7 @@ Enable Options
 
     * * ``Kokkos_ENABLE_CUDA_LAMBDA``
       * Activate experimental lambda features
-      * ``OFF``
```
crtrott, line 96: Why not: ``ON` (``OFF`` before 4.0)

dalg24: because it does not fit in the column

crtrott: ok

PR #363, opened 2023-04-21
Require reducers to define a monoid

masterleinad:
Fixes https://github.com/kokkos/kokkos/pull/6065#issuecomment-1518014632.

### DIFF
--- a/docs/source/API/core/builtinreducers/ReducerConcept.rst
+++ b/docs/source/API/core/builtinreducers/ReducerConcept.rst
@@ -105,7 +105,7 @@ The reducer is assumed to define a commutative monoid with respect to the value 
       return result;
     }
 
-is commutative and associative with identity element defined by ``reducer.init(el)``.
+is commutative and associative with identity element that can be set by calling ``reducer.init(el)``.
 
 
 Built-In Reducers
--- a/docs/source/API/core/builtinreducers/ReducerConcept.rst
+++ b/docs/source/API/core/builtinreducers/ReducerConcept.rst
@@ -95,7 +95,7 @@ Functions
 Requirements
 ~~~~~~~~~~~~
 
-The reducer is assumend to define a commutative monoid with respect to the value type it is used with, i.e., the binary operation
+The reducer is assumed to define a commutative monoid with respect to the value type it is used with, i.e., the binary operation
 
 .. code-block:: cpp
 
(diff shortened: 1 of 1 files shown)


fnrizzi (approved):

docs/source/API/core/builtinreducers/ReducerConcept.rst
```diff
@@ -92,6 +92,22 @@ Functions
...
+
+.. code-block:: cpp
+
+    value_type op(const value_type& val1, const value_type& val2) {
+      value_type result = val1;
+      reducer.join(result, val2);
+      return result;
+    }
+
+is commutative and associative with identity element defined by ``reducer.init(el)``.
```
dalg24, line 108: > with identity element defined by ``reducer.init(el)``.

Sounds a bit weird to me but not sure how to improve.

fnrizzi: per discussion: with identity element that can be set by calling ``reducer.init(el)``.

masterleinad:
```suggestion
is commutative and associative with identity element that can be set by calling ``reducer.init(el)``.
```

crtrott (approved):

docs/source/API/core/builtinreducers/ReducerConcept.rst
```diff
@@ -92,6 +92,22 @@ Functions
...
+.. code-block:: cpp
+
+    value_type op(const value_type& val1, const value_type& val2) {
+      value_type result = val1;
+      reducer.join(result, val2);
+      return result;
+    }
+
+is commutative and associative with identity element that can be set by calling ``reducer.init(el)``.
+
```
crtrott, line 109: We could also go with a "non-mathy" explanation:
```c++
auto foo(auto reducer, auto val1, auto val2) {
   auto result = val1;
   reducer.join(result, val2);
   return result;
}
```

returns `val` for both `foo(reducer, init, val)` and `foo(reducer, val, init)` where `init` is a value set by calling `reducer.init(init)`.

dalg24: I think your suggestion is a bit more appropriate for the wiki.

dalg24: Do you want to open a PR for it?

PR #364, opened 2023-04-24
stdalgos (1/10): team-level API, and rst update for release 4.2

fnrizzi:
Add team-level API to and convert to rst the following algorithms:
```
Kokkos_AdjacentFind.hpp
Kokkos_Count.hpp
Kokkos_CountIf.hpp
Kokkos_AllOf.hpp
Kokkos_AnyOf.hpp
Kokkos_NoneOf.hpp
```

NOTE: this follows the new structure for documenting the API, but does not use sphinx because the lack of properly aligning things make things quite unreadable.

Open individual html with Firefox (must be firefox or it won't display correctly) [Archive.zip](https://github.com/kokkos/kokkos-core-wiki/files/12736817/Archive.zip)

Example of what `count` rendered:

image

image


cedricchevalier19 (approved):
It renders correctly even on safari.

PR #369, opened 2023-05-01
Add documentation for Kokkos::printf

masterleinad:
Corresponds to https://github.com/kokkos/kokkos/pull/6083.

### DIFF
--- a/docs/source/API/core/printf.rst
+++ b/docs/source/API/core/printf.rst
@@ -1,0 +1,16 @@
+``Kokkos::printf``
+=================
+
+.. role:: cppkokkos(code)
+    :language: cppkokkos
+
+Defined in header ``<Kokkos_Core.hpp>``
+
+.. code-block:: cpp
+
+    template <typename... Args>
+    KOKKOS_FUNCTION void printf(const char* format, Args... args);
+
+Prints the data specified in ``format`` and ``args...`` to ``stdout``.
+The behavior is analogous to ``std::printf``, but the return type is ``void``
+to ensure a consistent behavior across backends.
--- a/docs/source/API/core/Utilities.rst
+++ b/docs/source/API/core/Utilities.rst
@@ -7,6 +7,7 @@ Utilities
    ./utilities/abort
    ./utilities/all
    ./utilities/complex
+   ./utilities/printf
    ./utilities/timer
    ./utilities/device_id
    ./utilities/num_threads


docs/source/API/core/printf.rst
```diff
@@ -0,0 +1,17 @@
...
+
+Defined in header ``<Kokkos_Core.hpp>``
+
+.. code-block:: cpp
+
+    template <typename... Args>
+    KOKKOS_FUNCTION int printf(const char* format, Args... args);
+
+For the Serial backend, ``Kokkos::printf`` is implemented in terms of ``std::printf`` printing the data specified in ``format`` and ``args...`` to ``stdout``.
+Returns the number of characters written if successful or a negative value if an error encountered.
```
ajpowelsnl, line 15: A few questions:

1) for serial implementations, `Kokkos::printf` just calls `std::printf`, and it is only in this case that formatted arguments will be printed to `stdout`. while in all other non-serial cases, you'll only have an `int` to `stdout`?

2) By "successful" vs. "error encountered," do you mean in the print statement itself?

3) For the non-serial cases, how useful is it to have a print statement return an `int` representing the characters that were written?  I'm assuming there's some difficulty in returning the characters themselves in the non-serial cases .... ?

docs/source/API/core/printf.rst
```diff
@@ -0,0 +1,17 @@
...
+
+The implementation for other backends emulates this behavior.
```
ajpowelsnl, line 17: By the "emulates this behavior," do you mean in all of the non-serial cases across all of the backends?

Final ask:  would it be possible for you to give a examples of serial and non-serial cases to see the difference in what you see in `stdout`?

docs/source/API/core/printf.rst
masterleinad: 1. No, they typically all print to stdout in the same way.
2. Yes, if, e.g., there is an encoding error, `std::printf` would return a negative value instead of the numbers of characters written. It's not quite clear what all of the device implementations would do in that (edge) case.
3. We could just not return anything but the idea is to wrap `std::printf` on the host and make a best effort to be close to that on the device.

docs/source/API/core/printf.rst
masterleinad: I don't have an example. `printf` behaves the same for all implementations right now to the bets of my knowledge but we don't necessarily want to promise that there might not be any deviations for device implementations.

ajpowelsnl: Group comment:

- Revisit once code PR is merged

PR #388, opened 2023-05-24
Threading Model

PhilMiller:
Document what semantics we actually have around use of multiple threads calling Kokkos

### DIFF
--- a/docs/source/multiple-threads.rst
+++ b/docs/source/multiple-threads.rst
@@ -1,0 +1,150 @@
+Use from Multiple Threads
+#########################
+
+This document describes how various Kokkos functionality can safely be called when using more than one thread of host execution.
+It also describes intentions for what Kokkos aspires to implement in support of known or anticipated use cases, with clear indications that they are not currently supported.
+Some discussion of implementation characteristics and performance expectations is included as well.
+
+We will use terms as defined by the C++ standard memory model. In that model, executions that include a data race have undefined behavior.
+Thus, use of Kokkos in ways that don't conform to the safety guidance given here may be considered to result in undefined behavior.
+
+In formal terms, we address two distinct issues here.
+*Concurrency* is the matter of what operations may safely execute in different threads. 
+*Linearizability* is the matter of how the results of concurrently-executed operations may map to potential observed orders in a hypothetical non-concurrent execution.
+
+Core
+====
+
+Fundamental Operations
+----------------------
+
+For purposes of multi-threaded execution, all free and member functions provided by Kokkos can be thought of as performing a sequence of one or more of the following *fundamental operations*
+
+
+* *Initialization*
+
+.. Not just Kokkos::init, but also whatever device-specific or thread-specific stuff we have Legion doing now
+
+* *Finalization*
+
+.. Ditto Initialization
+
+* *Memory Allocation*
+
+
+* *Memory Deallocation*
+
+
+* *Parallel Dispatch*
+  ``parallel_for``, ``parallel_reduce``, ``parallel_scan``, each of which occurs on a particular *Execution Space Instance*. If none is specified, Kokkos internally uses a default *Execution Space Instance*.
+
+* *Data Movement*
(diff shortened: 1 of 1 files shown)


docs/source/multiple-threads.rst
```diff
@@ -0,0 +1,150 @@
...
+For purposes of multi-threaded execution, all free and member functions provided by Kokkos can be thought of as performing a sequence of one or more of the following *fundamental operations*
+
+
+* *Initialization*
+
+.. Not just Kokkos::init, but also whatever device-specific or thread-specific stuff we have Legion doing now
+
+* *Finalization*
+
+.. Ditto Initialization
```
masterleinad, line 30: Backends can still only be initialized or finalized once. I'm not quite sure if it's worth mentioning initialization/finalization then. At the very least, we need to clarify what we mean here (execution space instance initialization/finalization maybe sensible).

docs/source/multiple-threads.rst
```diff
@@ -0,0 +1,150 @@
...
+
+..  Like what ``Kokkos::fence()`` does
+
+* *Local Synchronization*
+
+.. Like what we get from  ``execution_space_instance.fence()``, hypothetical ``is_running`` that returns ``false``
+
+* *Data Access*
+  ``View::operator()``, to memory that is accessible from the host.
+
```
masterleinad, line 55: Not quite sure if we want to promise anything about data access outside of kernels.

docs/source/multiple-threads.rst
```diff
@@ -0,0 +1,150 @@
...
+* Unmanaged Construction
+  Unmanaged construction performs no fundamental operations.
+* Destruction
+  Destruction occurs when the reference count of the underlying object to which a ``View`` refers falls to zero. Destruction performs a *Global Synchronization*, followed by a *Memory Deallocation*.
+  The *Global Synchronization* is performed to ensure that any preceding operations that may reference the memory owned by the underlying object *happen-before* that memory is released.
+  Any *Fundamental Operation* which accesses the ``View`` instance performed in a way that it does not have a *happens-before* relationship with Destruction results in undefined behavior.
+  (Note this could occur via improper concurrent use of a ``View`` variable shared between threads, one of which passes it to an operation, and another which reassigns it to reference a different instance)
+* Metadata Query
+* Element Access
+  Element Access performs a Data Access operation.
```
masterleinad, line 110: Not quite sure if we need these.

docs/source/multiple-threads.rst
```diff
@@ -0,0 +1,150 @@
...
+
+
+Backend-Specific Details
+------------------------
+
+.. Local or Global synchronizations below?
```
masterleinad, line 116: It might be enough to group backends into synchronous and asynchronous backends clarifying that kernels submitted by multiple kernels are serialized (if we decide to make that promise).

docs/source/multiple-threads.rst
```diff
@@ -0,0 +1,150 @@
...
+* ``OpenMP``
+  This backend is *synchronous* - there is implicitly a *Synchronization* after any *Memory Allocation*, *Memory Deallocation*, *Parallel Dispatch*, or *Data Movement* operation. 
+
+* ``Threads``
+  This backend is *synchronous* - there is implicitly a *Synchronization* after any *Memory Allocation*, *Memory Deallocation*, *Parallel Dispatch*, or *Data Movement* operation. 
+
+* ``CUDA`` and ``HIP``
+
+* ``HPX``
+
```
masterleinad, line 130: We should talk more about parallel dispatch and the behavior of independent threads (without a happens-before relationship between them) accessing the same data.
Possibly also clarifying were we promise that dispatch implies fences (linking to API for parallel_for, parallel_reduce, parallel_scan).

ajpowelsnl:
**TODO:**
- Champions: @masterleinad , @ajpowelsnl 
- Request reviews for:
  - #4385 
  - #6051
  - #6151
- Discuss Kokkos "thread safety" concepts with team
- Decide where to document "thread safety" behavior , using this PR as a starting point

Issue #378, opened 2023-05-10
Ensure that the documentation builds warning free when testing PRs


dalg24:
Our current CI builds do not check that PRs are not introducing warnings and it should.
Of course, the current HEAD does not build warning free so this will need fixing too.

ajpowelsnl:
@dalg24 -- is this a task for for the NextGen Analytics folks?

dalg24:
Sure

fnrizzi:
we own this

fnrizzi:
I think that what we should do is: 
- open issue with sphinx about the warning of duplicate declarations for overloads 
- in the meantime, figure out how to catch all warnings and detect if a PR intriduces new ones or if only existing ones are present which we already know are not dangerous

antoinemeyer5:
After analyzing the issue, we have 3 different warnings.

(1) "Invalid C++ declaration: Expected end of definition."
Number of warnings: 1

(2) "Inline emphasis start-string without end-string."
Number of warnings: 1

(3) "Duplicate C++ declaration, also defined at [...]"
Number of warnings: 91

But there is a subtlety for (3):
(3.1) the duplication may come from the same file or
(3.2) may be between different files 

For (3.1), the issue was in our `cppkokkos.py` script.
You can see the solution to fix this here: #396

So now if you merge this, it's only 51 warnings (3) left. 

So for (3.2), the warning is "Duplicate C++ declaration, also defined at API/core/builtinreducers/[file_name]" and in fact the issue in this directory comes from the way we've written the files.

So we're going to have to rewrite the files to actually declare classes with ``.. cppkokkos:struct:: [class_name]``

antoinemeyer5:
The aim of this issue is to ensure that documentation is built without errors or warnings when testing new PRs.

**The first step** was to fix all warnings so that the current HEAD branch could be built without warnings. This is the purpose of issue #420. Once validated and all PRs merged, the HEAD branch will build without warnings.

**The next step** in completely resolving this issue is to ensure that the CI fails if the PR code generates warnings. The easiest way is to use Sphinx's native options/flags. There is a flag that **turns warnings into errors**.

This solution avoids making changes in the CI and has the desired effect: if a warning is detected by Sphinx, compilation doesn't finish, returns a exit status 1 and the CI makes an error/failure. For more details, see PR #425

antoinemeyer5:
Updates:
- the last warning comes from #417 and is now fixed, wait for merge;
[...]

PR #396, opened 2023-06-07
Fix cppkokkos to avoid warnings

antoinemeyer5:
Related to #378

`kokkosinlinefunction` triggered "Duplicate C++ declaration, also defined...". 
This was due to cppkokkos.py missing to include this customization in a couple places. This PR fixes that so that these warnings are solved. 
NOTE: this does not fix the warnings all together, since to get rid of all of them we need #397


fnrizzi (approved):

cz4rs (approved):
Looks good to me.

Issue #397, opened 2023-06-07
all bultin reducers pages must use `.. cppkokkos:class` to avoid duplication warnings


fnrizzi:
@antoinemeyer5 and I found out that all the warnings related to the *builtin reducers* currently showing when building the doc are due to the fact that those pages do not use `.. cppkokkos:class` to define the class, they just list things in separate sections. 
So what happens is that sphinx does not have a way to "scope" those things within each class and thinks all functions (for example the method `join`) are duplicate. To avoid this, we need to make sure those pages use the class from sphinx explicitly. 
This issue is to track the progress for this: 

- [x] BAnd.rst : see #398 
- [ ] BOr.rst : see #401 
- [x] LAnd.rst : see #402
- [x] LOr.rst : see #404
- [x] Max.rst : see #405
- [x] MaxLoc.rst : see #406
- [x] Min.rst : see #407
- [x] MinLoc.rst : see #408
- [x] MinMax.rst : see #409
- [x] MinMaxLoc.rst : see #410
- [x] MinMaxLocScalar.rst : see #411
- [x] MinMaxScalar.rst : see #412
- [x] Prod.rst : see #413
- [x] Sum.rst : see #414

related to #378

antoinemeyer5:
@fnrizzi
Need clarification for #411 and #412.
Otherwise ready for review.

cz4rs:
PRs look good overall. 
`.. rubric:: Assignment Operator` might be redundant, but I don't think we should change it when fixing warnings.

PR #414, opened 2023-06-12
use proper sphinx class syntax for `Sum`

antoinemeyer5:
fix `Sum` according to #397 

| before | after |
| ------ | ----- | 
| image | image |
| image | image |
| image | image |

### DIFF
--- a/docs/source/API/core/builtinreducers/Sum.rst
+++ b/docs/source/API/core/builtinreducers/Sum.rst
@@ -1,9 +1,6 @@
 ``Sum``
 =======
 
-.. role::cpp(code)
-    :language: cpp
-
 .. role:: cppkokkos(code)
     :language: cppkokkos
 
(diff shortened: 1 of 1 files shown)


fnrizzi (approved):

docs/source/API/core/builtinreducers/Sum.rst
```diff
@@ -16,84 +13,94 @@ Usage
 
 .. code-block:: cpp
 
-    T result;
-    parallel_reduce(N,Functor,Sum<T,S>(result));
+   T result;
+   parallel_reduce(N,Functor,Sum<T,S>(result));
```
dalg24, line 17: Is the 3 spaces indentation intentional?

fnrizzi: i asked Antoine to do that to aling with the "c" of the code-block but we can change if needed

antoinemeyer5: Yes it's intentional, I stay tuned if you want to change

dalg24: I am not a big fan of it but I won't block because I don't know what is the common practice with restructured text and I don't want to add inconsistencies to the doc source.

cz4rs (approved):

Issue #415, opened 2023-06-13
alphabetical table should be converted to rst and fix all warnings to links


fnrizzi:
https://github.com/kokkos/kokkos-core-wiki/blob/main/docs/source/API/alphabetical.md this should be converted to rst 
and that should also fix all these warnings

```
/Users/fnrizzi/Desktop/work/kokkoscore/kokkos-core-wiki/docs/source/API/alphabetical.md:51: WARNING: 'myst' reference target not found: Cuda
/Users/fnrizzi/Desktop/work/kokkoscore/kokkos-core-wiki/docs/source/API/alphabetical.md:52: WARNING: 'myst' reference target not found: CudaSpace
/Users/fnrizzi/Desktop/work/kokkoscore/kokkos-core-wiki/docs/source/API/alphabetical.md:53: WARNING: 'myst' reference target not found: CudaUVMSpace
/Users/fnrizzi/Desktop/work/kokkoscore/kokkos-core-wiki/docs/source/API/alphabetical.md:54: WARNING: 'myst' reference target not found: CudaHostPinnedSpace
/Users/fnrizzi/Desktop/work/kokkoscore/kokkos-core-wiki/docs/source/API/alphabetical.md:57: WARNING: 'myst' reference target not found: ExecutionSpaceConcept
/Users/fnrizzi/Desktop/work/kokkoscore/kokkos-core-wiki/docs/source/API/alphabetical.md:60: WARNING: 'myst' reference target not found: HostSpace
/Users/fnrizzi/Desktop/work/kokkoscore/kokkos-core-wiki/docs/source/API/alphabetical.md:61: WARNING: 'myst' reference target not found: HPX
/Users/fnrizzi/Desktop/work/kokkoscore/kokkos-core-wiki/docs/source/API/alphabetical.md:65: WARNING: 'myst' reference target not found: is_array_layout
/Users/fnrizzi/Desktop/work/kokkoscore/kokkos-core-wiki/docs/source/API/alphabetical.md:66: WARNING: 'myst' reference target not found: is_execution_policy
/Users/fnrizzi/Desktop/work/kokkoscore/kokkos-core-wiki/docs/source/API/alphabetical.md:67: WARNING: 'myst' reference target not found: ExecutionSpaceConcept
/Users/fnrizzi/Desktop/work/kokkoscore/kokkos-core-wiki/docs/source/API/alphabetical.md:68: WARNING: 'myst' reference target not found: is_memory_space
[...]

PR #417, opened 2023-06-14
#415: API-alphabetical from md to rst

antoinemeyer5:
## Notes
- This PR is part of #378
- I have tested all the links.
- The file is identical to the original.

## Screenshots
image
image
image
image
image
image
image


fnrizzi:
> ine 12 and line 14: These are exactly the same lines. I have to remove the duplication, isn't it

yes

docs/source/API/alphabetical.rst
```diff
@@ -0,0 +1,231 @@
...
+| `SpaceAccessibility <core/SpaceAccessibility.html>`_                                 | `Core <core-index.html>`_ | `Spaces <core/Spaces.html>`_                                           | Facility to query accessibility rules between execution and memory spaces.                                                              |
++--------------------------------------------------------------------------------------+---------------------------+------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------+
+| `Subview Type <core/view/Subview_type.html>`_                                        | `Core <core-index.html>`_ | `View and related <core/View.html>`_                                   | Type of multi-dimensional array which is returned by the subview function                                                               |
```
crtrott, line 204: We can't do this: the keyword is `Subview` as stupid as it is.

crtrott (changes requested):
I don't think we can do the Subview thing this way. Otherwise looks good.

docs/source/API/alphabetical.rst
fnrizzi: looking into how to get rid of "Type" but avoid warning

docs/source/API/alphabetical.rst
```diff
@@ -54,6 +54,9 @@ Containers
 Core
 ----
 
+.. |SubviewType| replace:: Subview
+.. _SubviewType: core/view/Subview_type.html
```
fnrizzi, line 58: please write acomment to explain why this fixes it and your thinking behind this solution

antoinemeyer5:
There are two ways to create internal links with rst:

(1) simple link: ``` `Subview <core/view/Subview_type.html>`_ ```

The first solution (1) is simple, but **makes no difference between the link name and its display**. Internally, the link is called *Subview* and the display clickable by the user is *Subview*. In our case, there were several links with the same name, so we had a duplication warning.

(2) more customizable link:
```rst
.. declaration
.. |SubviewType| replace:: Subview
.. _SubviewType: core/view/Subview_type.html

.. use
|SubviewType|_
```

This second solution (2) **does the same thing and display**, but differentiates between:
- the internal name of the link (here *SubviewType*),
- the display of the link to the user (here *Subview*) and
- the destination of the link.

Thanks to this, you can converse with the desired display and avoid internal duplication. You must always be careful not to have another link in this file whose name is *SubviewType*, otherwise the duplication warning will return.

With this solution, it's much simpler to have internal link names with a certain shape and a different display. This separates rst's internal logic from the user's display.

fnrizzi:
ok makes sense to me

fnrizzi (approved):

nmm0:
BTW - In the future, we could potentially rely on the [index](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#special-names) since it's auto-generated

fnrizzi:
> BTW - In the future, we could potentially rely on the [index](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#special-names) since it's auto-generated

sounds a good idea

masterleinad (approved):

crtrott (approved):

Issue #433, opened 2023-07-05
verify that doc builds warnings-free when 417 is fixed


fnrizzi:
according to https://github.com/kokkos/kokkos-core-wiki/issues/420, once https://github.com/kokkos/kokkos-core-wiki/pull/417 is merged, the doc should build without warnings if using all the packages listed in [the requiremtns ](https://github.com/kokkos/kokkos-core-wiki/blob/main/build_requirements.txt). 

We should verify this is the case and report any problems or unexpected behavior.

antoinemeyer5:
I have checked @fnrizzi  and with #417, the doc build without any warnings.

fnrizzi:
i still get these:
```
/kokkos-core-wiki/docs/source/ProgrammingGuide/Atomic-Operations.md:85: WARNING: 'myst' cross-reference target not found: '../API/core/atomics/atomic_op_fetch.md' [myst.xref_missing]
/kokkos-core-wiki/docs/source/ProgrammingGuide/HierarchicalParallelism.md:266: WARNING: 'myst' cross-reference target not found: '../API/core/policies/TeamThreadRange.md' [myst.xref_missing]
/kokkos-core-wiki/docs/source/usecases/MDRangePolicy.md:91: WARNING: 'myst' cross-reference target not found: '../API/core/view/layoutLeft.md' [myst.xref_missing]
```

antoinemeyer5:
> i still get these:
> 
> ```
> /kokkos-core-wiki/docs/source/ProgrammingGuide/Atomic-Operations.md:85: WARNING: 'myst' cross-reference target not found: '../API/core/atomics/atomic_op_fetch.md' [myst.xref_missing]
> /kokkos-core-wiki/docs/source/ProgrammingGuide/HierarchicalParallelism.md:266: WARNING: 'myst' cross-reference target not found: '../API/core/policies/TeamThreadRange.md' [myst.xref_missing]
> /kokkos-core-wiki/docs/source/usecases/MDRangePolicy.md:91: WARNING: 'myst' cross-reference target not found: '../API/core/view/layoutLeft.md' [myst.xref_missing]
> ```

I confirm this error.

However, it does not appear in the CI here: https://github.com/kokkos/kokkos-core-wiki/actions/runs/5464124037/jobs/9945729283

I need to find out why I have missed mistakes. 

Do you have a idea or specific configuration @fnrizzi ?

fnrizzi:
[...]

PR #435, opened 2023-07-06
#433 : Fix last warnings to build warnings-free

antoinemeyer5:
We found 3 warnings here:

https://github.com/kokkos/kokkos-core-wiki/issues/433#issuecomment-1622198505

There are fixed with this PR


fnrizzi (changes requested):
this is changing a lot of lines because of the ending space. 
Please only make this about the fixes you made leaving everything else as it was

antoinemeyer5:
> this is changing a lot of lines because of the ending space. Please only make this about the fixes you made leaving everything else as it was

Done @fnrizzi

fnrizzi (approved):

PR #439, opened 2023-07-17
Fix warnings after #369 merge

antoinemeyer5:
#369 created two warnings:
- title underline too short
- `printf` is not referenced in any tables

I corrected them with this PR. 
It's all part of the issue: https://github.com/kokkos/kokkos-core-wiki/issues/378
We want to keep our build free of errors and warnings!


masterleinad (approved):

dalg24:
You say that the fact printf is not referenced anywhere causes a warning but this PR is only fixing the underline.
Did you mean to add a reference?

PR #443, opened 2023-07-21
New readme section : Requirements to build html page in virtual environments

antoinemeyer5:
Simple guide for indicate:
- how to set up from scratch a Python virtual environment locally on a machine,
- install in that virtual environment all the requirements for the wiki build and
- do the build in that "safe" environment

The aim is to help documentation creators to:
- be in a controlled environment where they can be sure of using the right tools, versions, etc,
- detect potential errors or warnings as early as possible

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # Python Virtual Environment
-/venv
+/env
 # PyCharm files
 /.idea
 # Rendered html files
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,5 +1,5 @@
 # Python Virtual Environment
-/venv
+/env
```
ajpowelsnl, line 2: CAVEAT: there might be limits on using this (PyCharm) IDE in certain environments.

PR #453, opened 2023-09-05
Update ParallelDispatch.md

aaronyicongfu:
KOKKOS_INLINE_FUNCTION is not the only decorator to mark device-callable functions. There is also KOKKOS_FUNCTION, which is equivalently useful and worth to be mentioned in this part of the tutorial.

### DIFF
--- a/docs/source/ProgrammingGuide/ParallelDispatch.md
+++ b/docs/source/ProgrammingGuide/ParallelDispatch.md
@@ -20,7 +20,7 @@ Important notes on syntax:
 
 A _functor_ is one way to define the body of a parallel loop. It is a class or struct<sup>1</sup> with a public `operator()` instance method. That method's arguments depend on both which parallel operation you want to execute (for, reduce, or scan), and on the loop's execution policy (e.g., range or team). For an example of a functor see the section in this chapter for each type of parallel operation. In the most common case of a [`parallel_for()`](../API/core/parallel-dispatch/parallel_for), it takes an integer argument which is the for loop's index. Other arguments are possible; see [Chapter 8 - Hierarchical Parallelism](HierarchicalParallelism).
 
-The `operator()` method must be const, and must be marked with the `KOKKOS_INLINE_FUNCTION` macro. If building with CUDA, this macro will mark your method as suitable for running on the CUDA device (as well as on the host). If not building with CUDA, the macro is unnecessary but harmless. Here is an example of the signature of such a method:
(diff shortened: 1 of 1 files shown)


dalg24 (approved):

docs/source/ProgrammingGuide/ParallelDispatch.md
```diff
@@ -20,7 +20,7 @@ Important notes on syntax:
...
 
-The `operator()` method must be const, and must be marked with the `KOKKOS_INLINE_FUNCTION` macro. If building with CUDA, this macro will mark your method as suitable for running on the CUDA device (as well as on the host). If not building with CUDA, the macro is unnecessary but harmless. Here is an example of the signature of such a method:
+The `operator()` method must be const, and must be marked with the `KOKKOS_FUNCTION` or `KOKKOS_INLINE_FUNCTION` macro. If building with CUDA, this macro will mark your method as suitable for running on the CUDA device (as well as on the host). If not building with CUDA, `KOKKOS_INLINE_FUNCTION` expands to `inline`, and `KOKKOS_FUNCTION` is unnecessary but harmless. Here is an example of the signature of such a method:
```
PhilMiller, line 23: Should this text also be more general than just CUDA, since we now support AMD and Intel accelerator hardware as well with similar requirements?

Maybe
```suggestion
The `operator()` method must be const, and must be marked with the `KOKKOS_FUNCTION` or `KOKKOS_INLINE_FUNCTION` macro. If building for a GPU backend, this macro will mark your method as suitable for running on the GPU device (as well as on the host). If not building for a GPU backend, `KOKKOS_INLINE_FUNCTION` expands to `inline`, and `KOKKOS_FUNCTION` is unnecessary but harmless. Here is an example of the signature of such a method:
```

dalg24: It is tricky because there is no annotation for SYCL (nor OpenMPTarget/OpenACC) but yes you are obviously correct this also applies to HIP.

ajpowelsnl: @dalg24 - should line 23 be re-worded to be more general, i.e., applying to CUDA and HIP, with all relevant caveats applied for SYCL (etc.)?

masterleinad: I don't think we need to mention which backends need the annotations and which don't. It should be enough to just say that you need the macros to support all the backends `Kokkos` has.

crtrott, line 23:
```suggestion
The `operator()` method must be const, and must be marked with the `KOKKOS_FUNCTION` or `KOKKOS_INLINE_FUNCTION` macro. This macro will mark your method as suitable for running on both accelerator devices and the host as necessary with some backends such as CUDA. If not building with any backends requiring markup, `KOKKOS_INLINE_FUNCTION` expands to `inline`, and `KOKKOS_FUNCTION` is unnecessary but harmless. Here is an example of the signature of such a method:
```

PR #461, opened 2023-10-09
Add taking the address of math constants to known issues

dalg24:
See https://github.com/kokkos/kokkos/pull/5897#issuecomment-1750690290

### DIFF
--- a/docs/source/API/core/numerics/mathematical-constants.rst
+++ b/docs/source/API/core/numerics/mathematical-constants.rst
@@ -40,8 +40,13 @@ All constants are defined in the ``Kokkos::numbers::`` namespace since version 4
 Notes
 -----
 
+.. _KnownIssues: ../../../known-issues.html#mathematical-constants
+
+.. |KnownIssues| replace:: known issues
+
 * The mathematical constants are available in ``Kokkos::Experimental::`` since Kokkos 3.6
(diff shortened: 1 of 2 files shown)


dalg24:
Not sure if we want to explicitly say NVCC does not support it.
They are also issues with OpenMPTarget and OpenACC.

Just in case someone else wants to explore 
https://godbolt.org/z/n83MWvz5j

docs/source/API/core/numerics/mathematical-constants.rst
```diff
@@ -40,8 +40,13 @@ All constants are defined in the ``Kokkos::numbers::`` namespace since version 4
 Notes
 -----
 
+.. _KnownIssues: ../../../known-issues.html#mathematical-constants
+
+.. |KnownIssues| replace:: known issues
+
 * The mathematical constants are available in ``Kokkos::Experimental::`` since Kokkos 3.6
 * They were "promoted" to the ``Kokkos::numbers`` namespace in 4.0
+* Avoid taking the address of mathematical constants in device code.  It is not supported by some toolchains, hence not portable.  (See |KnownIssues|_)
```
masterleinad, line 49:
```suggestion
* Passing mathematical constants by reference or taking their address in device code is not supported by some toolchains and hence not portable.  (See |KnownIssues|_)
```

masterleinad (approved):

PR #476, opened 2023-12-14
add an initial FAQ

### LINKED ISSUE #112: Start a Frequently Asked Questions page
We could start populating with things like "How do I join Kokkos on Slack?"

nmm0:
Closes #112

image

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -1,0 +1,8 @@
+FAQ
+###
+
+How do I join the Kokkos slack channel?
+  You can find the slack channel at `kokkosteam.slack.com <https://kokkosteam.slack.com>`_. Register a new account with your email. We automatically whitelist emails from most organizations, but if your email address is not whitelisted, you can contact the Kokkos maintainers (their emails are in the LICENSE file).
+
+How do I compile Kokkos with C++20 or C++23?
+  When configuring Kokkos with cmake, add the flag ``-DCMAKE_CXX_STANDARD=20`` (or ``23``). Ensure that the flag is also set for any downstream applications.
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -100,5 +100,6 @@ Website Content
    videolectures
    GitHub Repo <https://github.com/kokkos/kokkos>
    contributing
+   faq
    citation
    license


docs/source/faq.rst
```diff
@@ -0,0 +1,8 @@
+FAQ
+###
+
+How do I join the Kokkos slack channel?
+  You can find the slack channel at `kokkosteam.slack.com <https://kokkosteam.slack.com>`_. Register a new account with your email. We automatically whitelist emails from most organizations, but if your email address is not whitelisted, you can contact the Kokkos maintainers (their emails are in the LICENSE file).
+
+How do I compile Kokkos with C++20 or C++23?
+  When configuring Kokkos with cmake, add the flag ``-DCMAKE_CXX_STANDARD=20`` (or ``23``). Ensure that the flag is also set for any downstream applications.
```
ajpowelsnl, line 8: A general question:  Instead of FAQ, should we have a "Kokkos Quick Start?"

People I spoke to today said they wanted a "landing page" similar to  [Spack](https://spack.readthedocs.io/en/latest/)

Also, speaking of "landing page" we should try to "de-clutter" ours to be more aesthetic (again, using the Spack page as an example).

Going with the idea of a "Kokkos Quick Start," here's a guesstimate of main points to cover:

- Clone /download the project (link to gitHub project).  We should clarify the which branch / which stable release business in this step.  If you're a developer, you want to clone the project and use `develop`.  If you're purely a user, you'll want to download tar ball of "current stable release" (we need to fix up the project to reflect what that is)

- A few (CMake) source build "recipes" for a range of backends:  Serial, OpenMP, CUDA, HIP, SYCL, etc.  + C++ (17, 20, 23)

- Instructions for Spack builds, as an alternative

- Creating a Kokkos test program, w/ a `main.cpp` + Makefile 

- Join Kokkos Slack (link to our website instructions).  We should try to do as much cross-linking as possible.

ajpowelsnl (commented):
Maybe we should think about this a bit more, such that we end up with a clear guide on how to get up and running with Kokkos in not more than an hour (assuming basic programming proficiency).  A point that was made repeatedly today was that most people are not starting at the point of advanced C++ developer.  Instead, they're general scientific programmers.  We should calibrate our communication to that level.

docs/source/faq.rst
dalg24: We need a FAQ
Open a separate issue

dalg24 (approved):

PR #477, opened 2023-12-17
Document `Kokkos::kokkos_swap` and associated new public header

dalg24:
~~Pending on kokkos/kokkos#6678~~
Documenting kokkos/kokkos#6697

### DIFF
--- a/docs/source/API/core/utilities/swap.rst
+++ b/docs/source/API/core/utilities/swap.rst
@@ -1,0 +1,26 @@
+``Kokkos::swap``
+================
+
+.. role:: cppkokkos(code)
+    :language: cppkokkos
+
+Defined in header ``<Kokkos_Swap.hpp>``:sup:`(since 4.3)` which is included from ``<Kokkos_Core.hpp>``
+
+.. code-block:: cpp
+
+    template <class T>
+    KOKKOS_FUNCTION constexpr void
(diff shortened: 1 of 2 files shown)


docs/source/API/core/utilities/swap.rst
```diff
@@ -0,0 +1,26 @@
...
+    KOKKOS_FUNCTION constexpr void
+    swap(T2 (&a)[N], T2 (&b)[N]) noexcept(noexcept(*a, *b));  // (2) (since 4.3)
+
+
+1) Swaps the values ``a`` and ``b``. This overload does not participate in overload
+   resolution unless ``std::is_move_constructible_v<T> && std::is_move_assignable_v<T>``
+   is ``true``.
+
+2) Swaps the arrays ``a`` and ``b``. This overload does not participate in
+   overload resolution unless ``T2`` is swappable.
```
fnrizzi, line 26: can we say a little about what ``is_swappable`` is? maybe referring to the std https://en.cppreference.com/w/cpp/types/is_swappable?

dalg24: We would need to define our own which I was hoping we would not get into.
Also the way things are going (ADL issues) we will have to call Kokkos::saw something else...

crtrott (approved):

PR #481, opened 2024-01-09
Update `AoSoA` API in Cabana's use case

dalg24:
This page had never been updated since the device type template parameter was deprecated on the Cabana side.

@streeve please review

### DIFF
--- a/docs/source/usecases/SoA-and-AoSoA-with-Cabana.md
+++ b/docs/source/usecases/SoA-and-AoSoA-with-Cabana.md
@@ -55,7 +55,7 @@ class AoSoA;
 : The Kokkos memory space that carries information about where to allocate storage.
 
 `VectorLength`
-: The vector length for the structure of arrays (optional). If not specified, a default defined for each each memory space is used; this value will likely need to be modified for optimal performance.
+: The vector length for the structure of arrays (optional). If not specified, a default (defined per memory space) is used; this value may need to be modified for optimal performance.
 
 `MemoryTraits`
 : The Kokkos memory traits that tells who controls memory allocation and deallocation (optional).
--- a/docs/source/usecases/SoA-and-AoSoA-with-Cabana.md
+++ b/docs/source/usecases/SoA-and-AoSoA-with-Cabana.md
@@ -42,8 +42,7 @@ Defined in header [`<Cabana_AoSoA.hpp>`](https://github.com/ECP-copa/Cabana/blob
 
 ```C++
 template <class DataTypes, class MemorySpace,
-          int VectorLength = Impl::PerformanceTraits<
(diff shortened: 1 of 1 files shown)


docs/source/usecases/SoA-and-AoSoA-with-Cabana.md
```diff
@@ -52,8 +51,8 @@ class AoSoA;
 `DataTypes`
 : The types of the elements stored in the underlying `Cabana::SoA`s.
 
-`DeviceType`
-: The Kokkos device type that carries the information about where to execute code and where to allocate storage.
+`MemorySpace`
+: The Kokkos memory space that carries information about where to allocate storage.
 
 `VectorLength`
 : The vector length for the structure of arrays (optional).
```
streeve, line 58: Maybe worth adding that while we have a default for each backend, the user will likely need to modify this for optimal kernel performance

streeve (approved):
Thanks for fixing

docs/source/usecases/SoA-and-AoSoA-with-Cabana.md
dalg24: If you want to do "add a suggestion" I can merge into this PR.

streeve:
```suggestion
: The vector length for the structure of arrays (optional). If not specified, a default defined for each each memory space is used; this value will likely need to be modified for optimal performance.
```

docs/source/usecases/SoA-and-AoSoA-with-Cabana.md
```diff
@@ -52,11 +51,11 @@ class AoSoA;
...
 : The types of the elements stored in the underlying `Cabana::SoA`s.
 
-`DeviceType`
-: The Kokkos device type that carries the information about where to execute code and where to allocate storage.
+`MemorySpace`
+: The Kokkos memory space that carries information about where to allocate storage.
 
 `VectorLength`
-: The vector length for the structure of arrays (optional).
+: The vector length for the structure of arrays (optional). If not specified, a default defined for each each memory space is used; this value will likely need to be modified for optimal performance.
```
ldh4, line 58:
```suggestion
: The vector length for the structure of arrays (optional). If not specified, a default defined value is used for each memory space; this value will likely need to be modified for optimal performance.
```

streeve:
```suggestion
: The vector length for the structure of arrays (optional). If not specified, a default (defined per memory space) is used; this value may need to be modified for optimal performance.
```
Thanks for catching the mistake - I also softened the language since it's strongly kernel dependent

ldh4 (approved):

PR #487, opened 2024-01-31
parallel_reduce/scan: Clarify identity in case no reproducer was given

masterleinad:
Related to https://github.com/kokkos/kokkos/issues/6760.


docs/source/API/core/parallel-dispatch/parallel_reduce.rst
```diff
@@ -93,8 +93,8 @@ Requirements:
...
+* If the ``functor`` is a lambda, ``ReducerArgument`` must satisfy the ``Reducer`` concept or ``ReducerArgumentNonConst`` must be a POD type with ``operator +=`` and ``operator =`` or a ``Kokkos::View``.  In the latter case, a sum reduction is applied where the identity is assumed to be given by the default constructor of the value type (and not by ``reduction_identity```). If provided, the ``init``/ ``join``/ ``final`` member functions must not take a ``WorkTag`` argument even for tagged reductions.
+* If ``ExecPolicy`` is ``TeamThreadRange`` a "reducing" ``functor`` is not allowed and the ``ReducerArgument`` must satisfy the ``Reducer`` concept or ``ReducerArgumentNonConst`` must be a POD type with ``operator +=`` and ``operator =`` or a ``Kokkos::View``.  In the latter case, a sum reduction is applied where the identity is assumed to be given by the default constructor of the value type (and not by ``reduction_identity```).
```
cedricchevalier19, line 97: I think it can be clearer to mention the case where `reduction_identity` is used and when `default constructor value` is.
For a user perspective it is not even clear why `identity` is needed. Are you talking about initial value for the reduction or for some other operations in the algorithms?

masterleinad: The initial value is the identity (internally) so that we don't have to treat the first iteration as special.

masterleinad:
Waiting for @crtrott's proposal for a different behavior.

docs/source/API/core/parallel-dispatch/parallel_reduce.rst
ajpowelsnl: **Path forward:**
- accept as is, and document future behavior once refactor is in place

ajpowelsnl (approved):
**Post - Merge TODO:**
- Update once `parallel_reduce` , `parallel_scan` (identity) changes in place
- Create documentation issue once behavior updated, add a tag saying in `version xyz` reduction / scan identity can be used

docs/source/API/core/parallel-dispatch/parallel_scan.rst
```diff
@@ -49,7 +49,7 @@ Requirements:
 
   - The ``WorkTag`` free form of the operator is used if ``ExecPolicy`` is an ``IntegerType`` or ``ExecPolicy::work_tag`` is ``void``.
   - ``HandleType`` is an ``IntegerType`` if ``ExecPolicy`` is an ``IntegerType`` else it is ``ExecPolicy::member_type``.
-* The type ``ReturnType`` of the ``functor`` operator must be compatible with the ``ReturnType`` of the parallel_scan and must match the arguments of the ``init`` and ``join`` functions of the functor.
+* The type ``ReturnType`` of the ``functor`` operator must be compatible with the ``ReturnType`` of the parallel_scan and must match the arguments of the ``init`` and ``join`` functions of the functor if provided. If the functor doesn't have an ``init`` member function, it is assumed that the identity for the scan operation is given by the default constructor of the value type.
```
nmm0, line 52:
```suggestion
* The type ``ReturnType`` of the ``functor`` operator must be compatible with the ``ReturnType`` of the parallel_scan and must match the arguments of the ``init`` and ``join`` functions of the functor if provided. If the functor doesn't have an ``init`` member function, it is assumed that the identity for the scan operation is given by the default constructor of the value type (and not by ``reduction_identity```).
```

For consistency

nmm0 (approved):
minor addition

PR #501, opened 2024-02-14
Drop mentions of old CPU architectures

dalg24:
Update doc to reflect https://github.com/kokkos/kokkos/pull/6806


ajpowelsnl (approved):
Still too soon to remove KNL, SNB, Power8 ...?

PR #507, opened 2024-03-11
Document operator== for execution spaces

brian-kelley:
List ``operator==`` as non-member functionality of execution spaces. I had a case where I wanted to use this, but it wasn't listed here and I misread it as a private declaration in the actual code

### DIFF
--- a/docs/source/API/core/execution_spaces.rst
+++ b/docs/source/API/core/execution_spaces.rst
@@ -330,4 +330,4 @@ Non Member Facilities
 
 * ``template<class S1, class S2> struct SpaceAccessibility;``: typetraits to check whether two spaces are compatible (assignable, deep_copy-able, accessable). (see |KokkosSpaceAccessibility|_)
 
-* ``bool operator==(const S& lhs, const S& rhs)``: tests whether the two space instances (of the same type) are identical.
+* ``bool operator==(const execution_space& lhs, const execution_space& rhs)``: tests whether the two space instances (of the same type) are identical.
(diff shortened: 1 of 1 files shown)


masterleinad (approved):

docs/source/API/core/execution_spaces.rst
```diff
@@ -327,3 +329,5 @@ Non Member Facilities
 * ``template<class MS> struct is_execution_space;``: typetrait to check whether a class is a execution space.
 
 * ``template<class S1, class S2> struct SpaceAccessibility;``: typetraits to check whether two spaces are compatible (assignable, deep_copy-able, accessable). (see |KokkosSpaceAccessibility|_)
+
+* ``bool operator==(const S& lhs, const S& rhs)``: tests whether the two space instances (of the same type) are identical.
```
dalg24, line 333: What's `S`?  Should it read `ExecutionSpaceConcept`?

brian-kelley: Even though it's declared inside the ExecutionSpaceConcept class, it's a non-member friend function. So I was trying to follow the pattern of the other non-member features like the one above which uses S1,S2.

brian-kelley: I could change the type here to ``execution_space`` or ``ExecutionSpaceConcept`` or anything else if you prefer that

dalg24: I think `execution_space`

brian-kelley: Yeah, I actually like that better now too. I just pushed the change

PR #512, opened 2024-04-04
Improve backend selection documentation

dalg24:
As it says on the tin

### DIFF
--- a/docs/source/keywords.rst
+++ b/docs/source/keywords.rst
@@ -19,76 +19,86 @@ This page is organized in four sections:
 
 .. _keywords_backends:
 
-Kokkos backends
-===============
+Backend selection
+=================
+
+**Default State:**
+All backends are disabled by default.  This ensures you explicitly choose the
+backends you need for your specific hardware setup.
+
+**Enabling Backends:**
+You can enable backends by configuring with ``-DKokkos_ENABLE_<BACKEND>=ON``
+flag, where ``<BACKEND>`` is replaced with the specific backend you want to
+enable (e.g. ``-DKokkos_ENABLE_CUDA=ON`` for CUDA).
+
+**Restrictions:**
+  Mutual Exclusion: You can only have one device backend (e.g., CUDA, HIP,
+  SYCL) and one host parallel backend (e.g., OpenMP, C++ threads) enabled at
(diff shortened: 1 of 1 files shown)


masterleinad (commented):
We should also talk about the default execution space here and use that as (additional) motivation that we can't have two device backends.

dalg24:
> We should also talk about the default execution space here and use that as (additional) motivation that we can't have two device backends.

Not in the section that describes how you enable backends with CMake.

docs/source/keywords.rst
```diff
@@ -19,76 +19,86 @@ This page is organized in four sections:
...
+**Enabling Backends:**
+You can enable backends by configuring with ``-DKokkos_ENABLE_<BACKEND>=ON``
+flag, where ``<BACKEND>`` is replaced with the specific backend you want to
+enable (e.g. ``-DKokkos_ENABLE_CUDA=ON`` for CUDA).
+
+**Restrictions:**
+  Mutual Exclusion: You can only have one device backend (e.g., CUDA, HIP,
+  SYCL) and one host parallel backend (e.g., OpenMP, C++ threads) enabled at
+  the same time. This is because these backends manage parallelism in
+  potentially conflicting ways.
```
masterleinad, line 38:
```suggestion
  the same time. This is because these backends manage parallelism in
  potentially conflicting ways.
```
Does this argument hold for the device backends?

dalg24: Your suggestion is invalid (the code didn't change)

> Does this argument hold for the device backends?

Yes I think so

docs/source/keywords.rst
```diff
@@ -19,76 +19,86 @@ This page is organized in four sections:
...
 .. _keywords_backends:
 
-Kokkos backends
-===============
+Backend selection
+=================
+
+**Default State:**
+All backends are disabled by default.  This ensures you explicitly choose the
+backends you need for your specific hardware setup.
```
crtrott, line 27:
```suggestion
backends you need for your specific hardware setup.
If no backend is enabled explicitly, the Serial backend will be enabled.
```

masterleinad: Lines 44-45 already say that.

PR #525, opened 2024-05-15
Suggest to disable cudamallocasync for legacy device

cedricchevalier19:
Older Nvidia GPU can be tricky to use with Kokkos, write

### DIFF
--- a/docs/source/known-issues.rst
+++ b/docs/source/known-issues.rst
@@ -4,6 +4,20 @@ Known issues
 .. role:: cppkokkos(code)
     :language: cppkokkos
 
+CUDA
+====
+
+- When using a legacy NVIDIA gpu, Kokkos' initialization can report errors like:
+
+  .. code-block::
+
+     terminate called after throwing an instance of 'Kokkos::Experimental::CudaRawMemoryAllocationFailure'
+
+  A fix is to disable asynchronous memory allocations by adding the following to CMake arguments:
+
+  .. code-block::
+
+     -DKokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=OFF
 
 HIP
 ===
--- a/docs/source/known-issues.rst
+++ b/docs/source/known-issues.rst
@@ -7,7 +7,7 @@ Known issues
 CUDA
 ====
 
-- When using a legacy NVIDIA gpu, Kokkos' initialization can report errors like:
+- When using a legacy NVIDIA GPU, Kokkos' initialization can report errors like:
 
   .. code-block::
 
--- a/docs/source/known-issues.rst
+++ b/docs/source/known-issues.rst
@@ -7,7 +7,7 @@ Known issues
 CUDA
 ====
 
-- When using a legacy NVIDIA GPU, Kokkos' initialization can report errors like:
+- With older MPI versions and when using a legacy NVIDIA GPU the default allocation mechanism of Kokkos for `CudaSpace` can cause issues. For example MPI may crash with illegal memory accesses or Kokkos' initialization can report errors like:
 
   .. code-block::
 


docs/source/known-issues.rst
```diff
@@ -4,6 +4,20 @@ Known issues
 .. role:: cppkokkos(code)
     :language: cppkokkos
 
+CUDA
+====
+
+- When using a legacy NVIDIA gpu, Kokkos' initialization can report errors like:
```
dalg24, line 10: Do we have a good idea of what GPUs have this issue?  (If so it might be good to say "(e.g. pre Pascal architectures)" or whatnot.
Is it sensitive to the CUDA version?  (Is there something we could do on the Kokkos side?  The note is valuable regardless.)

masterleinad:
We know that this is an issue with older MPI backends. Did you see this in code that doesn't use MPI?

cedricchevalier19:
> We know that this is an issue with older MPI backends. Did you see this in code that doesn't use MPI?

Yes, we add issue without MPI.

docs/source/known-issues.rst
cedricchevalier19: I'll try to see. From what I remember, it was not easy to test on the Kokkos side, it depends both on the GPU and on how the system is configured (perhaps the issue with the other MPI backends is related to this).

The one report on slack was with sm_35 and cuda 11.8.  I may dive a little bit deeper in Nvidia documentation to see if from Pascal it is automatically safe or not.

cedricchevalier19: According to https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY__POOLS.html , what architectures are supported is not very clear.

However, we could do a better job at checking availability in Kokkos:
> Supported Platforms

> Whether or not a device supports the integrated stream ordered memory allocator may be queried by calling [cudaDeviceGetAttribute()](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1gb22e8256592b836df9a9cc36c9db7151) with the device attribute [cudaDevAttrMemoryPoolsSupported](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg49e2f8c2c0bd6fe264f2fc970912e5cd0876515d3b274cb0b40f8a7d913fb1aa).

cedricchevalier19: I think so.

masterleinad: See https://github.com/kokkos/kokkos/issues/7044.

docs/source/known-issues.rst
```diff
@@ -4,6 +4,20 @@ Known issues
 .. role:: cppkokkos(code)
     :language: cppkokkos
 
+CUDA
+====
+
+- When using a legacy NVIDIA GPU, Kokkos' initialization can report errors like:
```
ajpowelsnl, line 10: Is the term "legacy" here unambiguous?  In other words, maybe a simple table to show which Nvidia architectures / CUDA versions are / are not compatible with `cudamallocasync` and/or memory pools.

docs/source/known-issues.rst
```diff
@@ -4,6 +4,20 @@ Known issues
...
+
+  .. code-block::
+
+     terminate called after throwing an instance of 'Kokkos::Experimental::CudaRawMemoryAllocationFailure'
+
+  A fix is to disable asynchronous memory allocations by adding the following to CMake arguments:
+
+  .. code-block::
+
+     -DKokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=OFF
```
ajpowelsnl, line 20: In, this case, the `cudamallocasync` allocation mechanism is switched off.  Is it important to say what the alternative mechanism will be, as well as any side effects associated with using the alternative?

ajpowelsnl (commented):
Maybe a table to distill this information, instead of tasking users with interpreting ambiguous terms, such as "legacy."?

docs/source/known-issues.rst
cedricchevalier19: It is not straightforward to do a table as it depends on both the architecture and the driver/OS configuration.
Perhaps we should remove *legacy*, the goal is just to reference, if you see this error, try this to solve it.

crtrott, line 10:
```suggestion
- With older MPI versions and when using a legacy NVIDIA GPU the default allocation mechanism of Kokkos for `CudaSpace` can cause issues. For example MPI may crash with illegal memory accesses or Kokkos' initialization can report errors like:
```

masterleinad (approved):

PR #526, opened 2024-05-20
Reducer concept: Document final and mark init as optional

masterleinad:
We didn't mention the `final` member function in the reducer concept and providing an `init` member function is optional as well.

### DIFF
--- a/docs/source/API/core/builtinreducers/ReducerConcept.rst
+++ b/docs/source/API/core/builtinreducers/ReducerConcept.rst
@@ -33,14 +33,17 @@ Synopsis
             void join(value_type& dest, const value_type& src) const;
 
             KOKKOS_INLINE_FUNCTION
-            void init(value_type& val) const;
-
-            KOKKOS_INLINE_FUNCTION
             value_type& reference() const;
 
             KOKKOS_INLINE_FUNCTION
             result_view_type view() const;
 
+            //Optional
+            KOKKOS_INLINE_FUNCTION
+            void init(value_type& val) const;
+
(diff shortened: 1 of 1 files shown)


docs/source/API/core/builtinreducers/ReducerConcept.rst
```diff
@@ -82,7 +85,12 @@ Functions
 
 .. cppkokkos:kokkosinlinefunction:: void init(value_type& val) const;
 
-    * Initialize ``val`` with appropriate initial value. For example, 'Add' assigns ``val = 0;``, but Prod assigns ``val = 1;``
+    * Optional callback initializing ``val`` with appropriate initial value. For example, 'Add' assigns ``val = 0;``, but Prod assigns ``val = 1;``.
+      Defaults to calling the default constructor.
+
+.. cppkokkos:kokkosinlinefunction:: void final(value_type& val) const;
+
+    * Optional callback modifying the result ``val``. Defaults to a no-op.
```
dalg24, line 93: Should we be explicit and say "no-op (that is do nothing)" or you think it is obvious enough and we use that terminology other places?

dalg24 (commented):
Thanks for fixing

docs/source/API/core/builtinreducers/ReducerConcept.rst
masterleinad: We use it in a couple of other places so I think it's fine. I'm happy to change if requested, though.

dalg24 (approved):

PR #529, opened 2024-06-04
Consolidate virtual function pages

JBludau:
@dalg24 noticed that there are two pages on the topic of kokkos and virtual functions.
After a quick discussion on slack we propose to consolidate it and move it into the Programming Guide. We feel it relates more to the kokkos programming model than the use cases.
Nevertheless, this is open for discussion.

The consolidation basically consists of moving the special case (merged in https://github.com/kokkos/kokkos-core-wiki/pull/162) of the compiler optimization strategy circumventing Vtable lookup into the Programming Guide. With this done, the rest of the virtual functions usecase page was an exact copy of the page in the Programming Guide and is dropped.
Some resources are moved accordingly.


dalg24 (approved):
Thank you for fixing

PR #530, opened 2024-06-04
Remove unused DeviceVPointer and DeviceVTable pngs

JBludau:
While doing #529 I found that the pngs with virtual tables and virtual pointers on the device are currently unused.
```
grep -nr 'VirtualFunctions-DeviceVPointer.png' ./source | wc -l
       0
grep -nr 'VirtualFunctions-DeviceVTable.png' ./source | wc -l
       0
```
We should think if we want to keep them.

- If yes: maybe move them to `ProgrammingGuide/figures`
- if no: drop them (as proposed)


dalg24:
What do you recommend?  Do these figures convey information that is not captured by the other two images that are actually used or can we live w/o them?  Also maybe describe/attach them so it is clearer how the ones you remove differ from the ones that stay.

JBludau:
Your browser should show them if you go to `files changed` -> `tree dots` -> `view file`. 
But let's make it more convenient.

I am for deleting for two reasons:
- both of them are dangling files ... this is not going to get better.
- both of them have inconsistencies or errors when compared to the text and the existing images (I suspect that is why they are unused).

image
image


But:
- `VirtualFunctions-DeviceVPointer.png` (first image) can be added to the existing text, as the matching description is already there. And it would visualize the explanation. Nevertheless, the arrow should point from `vpointer` to `Host Vtable for Base` to be consistent with the rest (compare to [existing image](https://kokkos.org/kokkos-core-wiki/ProgrammingGuide/Kokkos-and-Virtual-Functions.html#v-tables-v-pointers-v-ery-annoying-with-gpus)). Furthermore, if the arrow points from `vpointer` to `Device Vtable` it can be used to visualize the Vpointer for [placement new on device](https://kokkos.org/kokkos-core-wiki/ProgrammingGuide/Kokkos-and-Virtual-Functions.html#how-to-fix-this).

- `VirtualFunctions-DeviceVTable.png` (second image) AFAIK the Vtable for the device is only created if there is at least one `placement new` call for the respective class on the device (pls correct me if compilers do this differently now). So it could be used to explain that once we do the [placement new on device](https://kokkos.org/kokkos-core-wiki/ProgrammingGuide/Kokkos-and-Virtual-Functions.html#how-to-fix-this) we have two tables now. But the image has an error: `Device Base::Foo()` in the last row of the middle column should be `Device Derived:Foo()` and the arrows in the lower half (the `Device` part) should be analogous to the `Host` part. 

I guess, the bottom line is: we could rework the page on virtual calls. The images could use the same names as the code examples and if we redo them they could visualize the explanation given in the text. I will propose something in a separate pr.

dalg24 (approved):
Thanks!

PR #535, opened 2024-06-10
Parallel dispatch, remove duplicates: one word and one clause

RL-S:
The first change is a correction, the other one is maybe a bit down to taste, but it seems easier to read and understand now.
I would guess that in the code, KOKKOS_LAMBDA is probably defined in terms of KOKKOS_INLINE_FUNCTION anyway, e.g. like this:

```c++
#if (defined KOKKOS_ENABLE_CUDA) || (defined KOKKOS_ENABLE_HIP) // or additional conditions, e.g. __CUDA_ARCH__
#define KOKKOS_INLINE_FUNCTION __host__ __device__ inline
#else
#define KOKKOS_INLINE_FUNCTION inline
#endif

#define KOKKOS_LAMBDA KOKKOS_INLINE_FUNCTION [=]
```

If so, then the _special meaning_ is referring to the same thing. If not, then maybe a clarification would be good.


docs/source/ProgrammingGuide/ParallelDispatch.md
```diff
@@ -38,7 +38,7 @@ The 2011 version of the C++ standard ("C++11") provides a new language construct
...
 If CUDA is disabled, this just turns into the usual capture-by-value clause `[=]`. That captures variables from the surrounding scope by value. Do NOT capture them by reference! If CUDA is enabled, this macro may have a special definition
-that makes the lambda work correctly with CUDA. Compare to the `KOKKOS_INLINE_FUNCTION` macro, which has a special meaning if CUDA is enabled. If you do not plan to build with CUDA, you may use `[=]` explicitly, but we find using the macro easier than remembering the capture clause syntax.
+that makes the lambda work correctly with CUDA, similar to the `KOKKOS_INLINE_FUNCTION` macro. If you do not plan to build with CUDA, you may use `[=]` explicitly, but we find using the macro easier than remembering the capture clause syntax.
```
masterleinad, line 41: I would just say:
> Similar to the `KOKKOS_INLINE_FUNCTION` and `KOKKO_FUNCTION` macro, `KOKKOS_LAMBDA` adds backend-specific annotations for portability.

or similar in an attempt to avoid talking about specific backends.

RL-S: I don't mind that (except the `S` in `KOKKOS_FUNCTION` is missing). But I'm also not opposed to mentioning that it's about the `__host__` and `__device__` specifiers. Because that's valuable information for people who want to migrate their code from CUDA/HIP.

ajpowelsnl:
Hi @RL-S , @masterleinad -- I do hate to be a PR party pooper, but for now, we are 110% focused on revising the [API guide](https://kokkos.org/kokkos-core-wiki/), and not the Programming guide.

In the API Guide, our goals are (in loose order):

- fill in any gaps
- make sure documentation is correct with respect to the latest release
- correct errors
- improve / extend examples

If you would like help getting up and running with the API Guide repo, please LMK, and I'll be happy to help.  Again, I am sorry to be a PR party pooper.

docs/source/ProgrammingGuide/ParallelDispatch.md
JBludau: Hmm I am with @masterleinad and would avoid spelling out anything about specifics for certain backends. Apart from possible changes to `KOKKOS_LAMBDA` in the future, most places in the doc refrain from using specifics.
I actually don't like the Cuda example specifics here ...  

For me `KOKKOS_LAMBDA` just marks a lambda to be usable with Kokkos. For other lambdas there is simply no guarantee

RL-S: Just want to mention that I didn't introduce the reference to CUDA/HIP. In both places with changes, it existed already:
1. "For some backends (such as CUDA and HIP) ..."
2. "If CUDA is disabled, [...]. If CUDA is enabled, [it] makes the lambda work correctly with CUDA [...]"
So while I still uphold that it _can help_ when migrating _and doesn't harm_ otherwise, it's simply not something introduced by this commit.

Btw, [API Core - parallel_for](https://kokkos.org/kokkos-core-wiki/API/core/parallel-dispatch/parallel_for.html#examples) also mentions how the macros relate. Maybe that page should link here and/or vice versa.

JBludau: Hmm I guess my point is: If you are porting something from whatever to Kokkos, you are basically porting it to a new `programming model`. This model is translated to the e.g. Cuda's programming model by kokkos. Nevertheless, this does not imply, that things in the Kokkos programming model must have an equivalent in all backends that kokkos can translate to. And thinking e.g. Cuda in the context of kokkos might lead to false assumptions as it is not the same programming model.
Thus I would favor just stating what you get from the annotation -  plane and simple.

RL-S: I'm not sure I agree with that assessment. Elsewhere in the docs it also states that backend-specific code still compiles. It's a strength of Kokkos, imo, that it often behaves like an abstraction on top of the different backends, because that allows step-by-step migration, as well as even backend-specific code sections when properly `#ifdef`ed.

But I think that's beside the point of this commit, which doesn't make any change to the philosophy of the wiki page. I think this commit improves the page, and I'm very okay with it being a temporary improvement, if you rather want to fully remove any reference to backends in the future. But let's do these things one at a time.

docs/source/ProgrammingGuide/ParallelDispatch.md
```diff
@@ -38,7 +38,7 @@ The 2011 version of the C++ standard ("C++11") provides a new language construct
 
 By default, lambdas capture nothing (as the default capture specifier `[]` specifies). This is not likely to be useful, since [`parallel_for()`](../API/core/parallel-dispatch/parallel_for) generally works by its side effects. Thus, we recommend using the ``capture by value'' specifier `[=]` by default. You may also explicitly specify variables to capture, but they must be captured by value. We prefer that for the outermost level of parallelism (see [Chapter 8](HierarchicalParallelism)), you use the `KOKKOS_LAMBDA` macro instead of the capture clause.
 If CUDA is disabled, this just turns into the usual capture-by-value clause `[=]`. That captures variables from the surrounding scope by value. Do NOT capture them by reference! If CUDA is enabled, this macro may have a special definition
```
crtrott, line 40:
```suggestion
For some backends, this just turns into the usual capture-by-value clause `[=]`. That captures variables from the surrounding scope by value. Do NOT capture them by reference! For other backends (e.g. CUDA and HIP), this macro may have a special definition
```

docs/source/ProgrammingGuide/ParallelDispatch.md
crtrott, line 41:
```suggestion
that makes the lambda work correctly with some execution spaces, same as the `KOKKOS_INLINE_FUNCTION` macro. 
```

docs/source/ProgrammingGuide/ParallelDispatch.md
```diff
@@ -38,7 +38,7 @@ The 2011 version of the C++ standard ("C++11") provides a new language construct
 
 By default, lambdas capture nothing (as the default capture specifier `[]` specifies). This is not likely to be useful, since [`parallel_for()`](../API/core/parallel-dispatch/parallel_for) generally works by its side effects. Thus, we recommend using the ``capture by value'' specifier `[=]` by default. You may also explicitly specify variables to capture, but they must be captured by value. We prefer that for the outermost level of parallelism (see [Chapter 8](HierarchicalParallelism)), you use the `KOKKOS_LAMBDA` macro instead of the capture clause.
```
crtrott, line 39:
```suggestion
By default, lambdas capture nothing (as the default capture specifier `[]` specifies). This is not likely to be useful, since [`parallel_for()`](../API/core/parallel-dispatch/parallel_for) generally works by its side effects. Because Kokkos reserves the right to make copies of the closure, and its operations are potentially asynchronous users must ``capture by value'' to be semantically correct. We recommend doing so via the KOKKOS_LAMBDA macro for the outermost level of parallelism (see [Chapter 8](HierarchicalParallelism)).
```

crtrott (approved):

PR #539, opened 2024-06-11
create_mirror and create_mirror_view: clarified difference

RL-S:
Added a paragraph to "Usage", before the code example, and added comments to the code example.

I was thinking about changing the variable names in the second half of the code example from `host_mirror_space*` to `space_mirror*`, as now they don't *have* to be in host space, if another space is provided. Though I'm not yet familiar enough with Kokkos to know whether `ExecName()` is pseudo code or evaluates to the default execution space or something like that.

### DIFF
--- a/docs/source/API/core/view/create_mirror.rst
+++ b/docs/source/API/core/view/create_mirror.rst
@@ -16,7 +16,7 @@ Usage
 -----
 
 The key difference between ``create_mirror`` and ``create_mirror_view`` is the following: ``create_mirror`` `always` allocates new memory in the specified space (shown below for host space), while ``create_mirror_view`` only allocates memory if the View to be mirrored (``a_view``) is not already accessible from the specified space, and otherwise simply returns ``a_view``.
-Use ``create_mirror_view`` when the mirror is solely used for providing access in different execution spaces, and use ``create_mirror`` if you need the data to be independent, e.g. for having an previous and a updated version of the data.
+Use ``create_mirror_view`` when the mirror is solely used for providing access in different execution spaces, and use ``create_mirror`` if you need the data to be independent, e.g. for having a previous and an updated version of the data.
 
 .. code-block:: cpp
 
--- a/docs/source/API/core/view/create_mirror.rst
+++ b/docs/source/API/core/view/create_mirror.rst
@@ -26,9 +26,9 @@ Use ``create_mirror_view`` when you want to skip the memory allocation and copy 
     // host_mirror_view may point to the same memory as a_view, if a_view is host-accessible
     auto host_mirror_view = create_mirror_view(a_view);
 
-    // You can specify the execution space from which the mirror view must be accessible
-    auto host_mirror_space = create_mirror(ExecSpace(),a_view);
-    auto host_mirror_view_space = create_mirror_view(ExecSpace(),a_view);
+    // You can specify the space from which the mirror view must be accessible
+    auto mirror = create_mirror(memory_space_instance, a_view);
+    auto mirror_view = create_mirror_view(memory_space_instance, a_view);
 
 
 Description
--- a/docs/source/API/core/view/create_mirror.rst
+++ b/docs/source/API/core/view/create_mirror.rst
@@ -15,13 +15,21 @@ A common desired use case is to have a memory allocation in GPU memory and an id
 Usage
 -----
 
+The key difference between ``create_mirror`` and ``create_mirror_view`` is the following: ``create_mirror`` `always` allocates new memory in the specified space (shown below for host space), while ``create_mirror_view`` only allocates memory if the View to be mirrored (``a_view``) is not already accessible from the specified space, and otherwise simply returns ``a_view``.
+Use ``create_mirror_view`` when you want to skip the memory allocation and copy within the same execution space, and use ``create_mirror`` if you need the data to be independent, e.g. for overlapping computation and I/O.
+
 .. code-block:: cpp
 
+    // Both host_mirror and host_mirror_view have the correct properties for deepCopy from/to a_view
+    // host_mirror is guaranteed to have separately allocated memory from a_view
     auto host_mirror = create_mirror(a_view);
+    // host_mirror_view may point to the same memory as a_view, if a_view is host-accessible
     auto host_mirror_view = create_mirror_view(a_view);
 
+    // You can specify the execution space from which the mirror view must be accessible
     auto host_mirror_space = create_mirror(ExecSpace(),a_view);
     auto host_mirror_view_space = create_mirror_view(ExecSpace(),a_view);
+
 
 Description
 -----------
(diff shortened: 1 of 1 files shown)


docs/source/API/core/view/create_mirror.rst
```diff
@@ -15,14 +15,22 @@ A common desired use case is to have a memory allocation in GPU memory and an id
 Usage
 -----
 
+The key difference between ``create_mirror`` and ``create_mirror_view`` is the following: ``create_mirror`` `always` allocates new memory in the specified space (shown below for host space), while ``create_mirror_view`` only allocates memory if the View to be mirrored (``a_view``) is not already accessible from the specified space, and otherwise simply returns ``a_view``.
+Use ``create_mirror_view`` when you want to skip the memory allocation and copy within the same execution space, and use ``create_mirror`` if you need the data to be independent, e.g. for overlapping computation and I/O.
```
masterleinad, line 19: In my opinion, this fits better in the block above. Also, the last sentence seems redundant, in particular, since the interface is primarily about memory spaces and not execution spaces.

docs/source/API/core/view/create_mirror.rst
```diff
@@ -15,14 +15,22 @@ A common desired use case is to have a memory allocation in GPU memory and an id
...
 
+    // Both host_mirror and host_mirror_view have the correct properties for deepCopy from/to a_view
+    // host_mirror is guaranteed to have separately allocated memory from a_view
     auto host_mirror = create_mirror(a_view);
+    // host_mirror_view may point to the same memory as a_view, if a_view is host-accessible
     auto host_mirror_view = create_mirror_view(a_view);
 
+    // You can specify the execution space from which the mirror view must be accessible
     auto host_mirror_space = create_mirror(ExecSpace(),a_view);
     auto host_mirror_view_space = create_mirror_view(ExecSpace(),a_view);
```
masterleinad, line 31: This should really not be called `host_mirror_view_space`. I would rather have an example that provides a memory space instance since there is also the API using `view_alloc` that allows providing an execution space instance that is used for initialization and copying.
```suggestion
    // You can specify the space from which the mirror view must be accessible
    auto mirror = create_mirror(memory_space_instance, a_view);
    auto mirror_view = create_mirror_view(memory_space_instance, a_view);
```

docs/source/API/core/view/create_mirror.rst
RL-S: Feel free to move it around. I think an illustrative example (the last sentence) is very helpful, so some form of it should remain in my opinion.
For execution vs memory space: The example under "Usage" already had a parameter called "ExecSpace", so I just stuck to that. But yeah, I wasn't sure about that part.

RL-S: In the sentence
"Use ``create_mirror_view`` when you want to skip the memory allocation and copy within the same execution space, and use ``create_mirror`` if you need the data to be independent, e.g. for overlapping computation and I/O."
we could remove "execution". That would now also fit with the example code.

JBludau: Hmm ... stating that a copy is avoided implies, that a copy is always required.

JBludau: I would drop the sentence with the examples. It sounds too much like a recommendation tbh.

RL-S: That sentence was a bit tricky. But while it's not airtight, I think it gets the message across with good brevity and clarity.
It is written like a recommendation, and I think that's exactly what is needed when functions with this level of similarity are provided. But we could also write something like
"Using ``create_mirror_view`` allows skipping the memory allocation and potential copies within the same execution space, while ``create_mirror`` can be used if you need the data to be independent, e.g. for overlapping computation and I/O."

Feel free to make a different suggestion. But leaving out the example sounds like the worst option to me.

docs/source/API/core/view/create_mirror.rst
```diff
@@ -15,13 +15,21 @@ A common desired use case is to have a memory allocation in GPU memory and an id
 Usage
 -----
 
+The key difference between ``create_mirror`` and ``create_mirror_view`` is the following: ``create_mirror`` `always` allocates new memory in the specified space (shown below for host space), while ``create_mirror_view`` only allocates memory if the View to be mirrored (``a_view``) is not already accessible from the specified space, and otherwise simply returns ``a_view``.
+Use ``create_mirror_view`` when you want to skip the memory allocation and copy within the same execution space, and use ``create_mirror`` if you need the data to be independent, e.g. for overlapping computation and I/O.
+
 .. code-block:: cpp
 
+    // Both host_mirror and host_mirror_view have the correct properties for deepCopy from/to a_view
```
crtrott, line 23:
```suggestion
    // Both host_mirror and host_mirror_view have the correct properties for deep_copy from/to a_view
```

docs/source/API/core/view/create_mirror.rst
```diff
@@ -15,13 +15,21 @@ A common desired use case is to have a memory allocation in GPU memory and an id
 Usage
 -----
 
+The key difference between ``create_mirror`` and ``create_mirror_view`` is the following: ``create_mirror`` `always` allocates new memory in the specified space (shown below for host space), while ``create_mirror_view`` only allocates memory if the View to be mirrored (``a_view``) is not already accessible from the specified space, and otherwise simply returns ``a_view``.
+Use ``create_mirror_view`` when you want to skip the memory allocation and copy within the same execution space, and use ``create_mirror`` if you need the data to be independent, e.g. for overlapping computation and I/O.
```
crtrott, line 19:
```suggestion
Use ``create_mirror_view`` when the mirror is solely used for providing access in different execution spaces, and use ``create_mirror`` if you need the data to be independent, e.g. for having an previous and a updated version of the data.
```

masterleinad (approved):

docs/source/API/core/view/create_mirror.rst
```diff
@@ -15,13 +15,21 @@ A common desired use case is to have a memory allocation in GPU memory and an id
 Usage
 -----
 
+The key difference between ``create_mirror`` and ``create_mirror_view`` is the following: ``create_mirror`` `always` allocates new memory in the specified space (shown below for host space), while ``create_mirror_view`` only allocates memory if the View to be mirrored (``a_view``) is not already accessible from the specified space, and otherwise simply returns ``a_view``.
+Use ``create_mirror_view`` when the mirror is solely used for providing access in different execution spaces, and use ``create_mirror`` if you need the data to be independent, e.g. for having an previous and a updated version of the data.
```
ldh4, line 19:
```suggestion
Use ``create_mirror_view`` when the mirror is solely used for providing access in different execution spaces, and use ``create_mirror`` if you need the data to be independent, e.g. for having a previous and an updated version of the data.
```

PR #544, opened 2024-06-18
Correct display of superscripts

rb214678:
Two superscripts were missing a backslash-escaped whitespace (see https://documatt.com/restructuredtext-reference/element/superscript.html) to be correctly displayed. Add a newline between the blocks explaining the superscripts for readability.


masterleinad:
https://github.com/kokkos/kokkos-core-wiki/pull/531 fixes the same issue.

rb214678:
My bad, Sorry I did not even looked at currently opened PR. I propose to edit my PR to keep the part "Add a newline between the blocks explaining the superscripts for readability." which is not covered in #531.

PR #551, opened 2024-07-09
Improve documentation for virtual functions

pzehner:
This PR aims to deeply rework the documentation relative to virtual functions:

- Removed mixed Cuda/Kokkos code;
- Start from a serial example case;
- Improve `kokkos_malloc` code: the documentation used `Kokkos::kokkos_malloc` in a way that couldn't compile. A `static_cast` was needed. In the same Slack discussion, it was suggested to use a more precise formulation (separate allocated memory and actual object);
- Make the code more portable: only resort to Kokkos features;
- Make examples easier to build: most examples were given as "pseudo-code", this was changed to actual code, with comments to explicit what is done;
- List backends for which virtual functions in parallel region work;
- Display a warning about the use of virtual functions in general at the beginning of the page; and
- Fix typos.


ajpowelsnl (commented):
This entry is rather confusing, not least of all because there's co-mingling of Kokkos and CUDA code.  If the goal is to provide an explanation of virtual functions in Kokkos (and how to access them correctly from host and device), this entry takes the long way around the proverbial barn.

A few more suggestions:

- Point users to a couple quality C++ resources on virtual functions (you already have one)
- Remove "hybrid" code
- Remove code that is intentionally incorrect
- Focus on the correct Kokkos code (making sure the example will compile and run)
- Brief explanation of why this works (i.e., Virtual Tables, Virtual pointers)
- Maybe mention "pitfalls" to alert the reader to the fact that derived class methods on host and device, respectively, have strict access patterns, where one (host) cannot access the other (device)

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -4,76 +4,85 @@ Due to oddities of GPU programming, the use of virtual functions in Kokkos paral
...
 
-To one where we construct on the device using a technique called `placement new`
+to one where we construct on the device using a technique called *placement new*
 
 ```c++
-Base* deviceInstance; // cudaMalloc'd to allocate it
-Kokkos::parallel_for(... {
-  new((Derived*)deviceInstance) Derived(); // construct an instance in the place, the pointer deviceInstance points to
+Base* deviceInstance; // declare
+cudaMalloc((void**)&deviceInstance, sizeof(Derived)); // alocate on device
```
cedricchevalier19, line 78:
```suggestion
cudaMalloc((void**)&deviceInstance, sizeof(Derived)); // allocate on device
```

cedricchevalier19:
> This entry is rather confusing, not least of all because there's co-mingling of Kokkos and CUDA code. If the goal is to provide an explanation of virtual functions in Kokkos (and how to access them correctly from host and device), this entry takes the long way around the proverbial barn.

I think @pzehner's goal was to update the page to make the examples work. However, I agree with @ajpowelsnl that this page is quite complicated to read, and I do not think the middle section with explicit Cuda code is very appealing to a Kokkos user.

More generally, do we want to advertise using virtual functions on GPU?

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -82,87 +91,125 @@ This code is extremely ugly, but leads to functional virtual function calls on t
...
-    {
-      auto derivedPtr = std::make_shared<Derived>();
-      derivedPtr->apply();
-      Kokkos::fence();
-    }
-    Kokkos::finalize();
+  Kokkos::ScopeGuard kokkos(argc, argv);
+  auto derivedPtr = std::make_shared<Derived>();
+  derivedPtr->apply();
+  Kokkos::fence();
```
JBludau, line 207: why use the `ScopeGuard` here?

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -82,87 +91,125 @@ This code is extremely ugly, but leads to functional virtual function calls on t
...
+  deviceInstance->~Base(); // destroy on device
+});
+Kokkos::fence("Wait for destroy");
+Kokkos::kokkos_free(deviceInstanceMemory); // free
+```
+
+We replaced `cudaMalloc()` by `Kokkos::kokkos_malloc()` and introduced a distinction between the memory that the instance uses, and the actual instance object.
+Since the kernel does not have a return type, we use a static cast to associate the later to the former.
+Then, the object is destroyed, and its memory freed.
+
```
JBludau, line 122: I guess if we put the kokkos version we should do it all around and get rid of the cuda version

JBludau:
> More generally, do we want to advertise using virtual functions on GPU?

I would prefer not advising the use of virtual functions on GPUs. But I had the feeling, the page did show how it can be done while expressing the difficulties that this brings to the table in quite a neutral tone

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -4,76 +4,85 @@ Due to oddities of GPU programming, the use of virtual functions in Kokkos paral
...
-Now that we know what the compiler is doing to implement virtual functions, we'll look at why it doesn't work with GPU's
+Now that we know what the compiler is doing to implement virtual functions, we'll look at why it doesn't work with GPU's.
 
-Credit: the content of this section is adapted from Pablo Arias [here](https://pabloariasal.github.io/2017/06/10/understanding-virtual-tables/)
+Credit: the content of this section is adapted from Pablo Arias [here](https://pabloariasal.github.io/2017/06/10/understanding-virtual-tables/).
 
 ## Then why doesn't my code work?
 
-The reason the intro code might break is that when dealing with GPU-compatible classes with virtual functions, there isn't one Vtable, but two. The first holds the host versions of the virtual functions, while the second holds the device functions. 
+The reason the intro code might break is that when dealing with GPU-compatible classes with virtual functions, there isn't one Vtable, but two. The first holds the host version of the virtual functions, while the second holds the device functions.
```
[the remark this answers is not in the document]
pzehner: This is not part of this PR.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
pzehner: `ScopeGuard` should be preferred over `initialize`/`finalize` as it ensures that `finalize` is called no matter what (RAII).

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
pzehner: I'm not sure to get what you mean.

JBludau: I meant update all the code sections to use kokkos semantics and get rid of the cuda code.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
JBludau: hmm AFAIK we try to avoid `ScopeGuard` in the examples, both in the tutorials as well as in the documentation. I do agree, that it would not make much of a difference in this example, but I would favor consistency

pzehner:
> This entry is rather confusing, not least of all because there's co-mingling of Kokkos and CUDA code. If the goal is to provide an explanation of virtual functions in Kokkos (and how to access them correctly from host and device), this entry takes the long way around the proverbial barn.

@ajpowelsnl As stated in the description of this PR, this contribution is an improvement of the existing documentation page. Its pedagogical approach (i.e. presenting the problem with a failing mixed CUDA-Kokkos code, then iterating to a more Kokkos solution) wasn't changed. Many remarks that you made about this PR are more remarks about the original page. My objectives were clearly stated at the beginning of this discussion.

Now we're here, we can discuss about changing the page more deeply.

> Point users to a couple quality C++ resources on virtual functions (you already have one)

I don't think more resources would be necessary. If a developer even accesses this documentation page, it's reasonable to consider they know what a virtual function is. If you have any interesting reference, feel free to tell me, however.

> - Remove "hybrid" code

The approach of the original article is indeed weird: nobody mixes CUDA and Kokkos code directly like that. I think it would make more sense to start from a plain serial code and get a portable Kokkos code. Or from a full CUDA code and get a Kokkos code (which is maybe the original motivation for this page).

> - Remove code that is intentionally incorrect

Naive approaches could be just described, instead of being presented in a code block.

> - Focus on the correct Kokkos code (making sure the example will compile and run)

This PR makes the Kokkos code at least buildable and proposes a full Kokkos approach.

> - Brief explanation of why this works (i.e., Virtual Tables, Virtual pointers)

I believe the page already explains it.

> Maybe mention "pitfalls" to alert the reader to the fact that derived class methods on host and device, respectively, have strict access patterns, where one (host) cannot access the other (device)

This could be improved in text, indeed.

The illustrations, if I got them right, already show what is possible and not.

> More generally, do we want to advertise using virtual functions on GPU?

> I would prefer not advising the use of virtual functions on GPUs. But I had the feeling, the page did show how it can be done while expressing the difficulties that this brings to the table in quite a neutral tone

I think we should warn the reader at the very beginning of the article that using virtual functions in parallelized regions is a bad idea in the first place. It is known to have bad performance, it's inconvenient and ugly to use on GPU... @masterleinad confirmed me on Slack that this practice is discouraged.

I'll add such a warning at the beginning of the article.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
pzehner: Ah, yes indeed.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -4,76 +4,85 @@ Due to oddities of GPU programming, the use of virtual functions in Kokkos paral
...
 };
 
-Base* hostClassInstance = new Derived();
-Base* deviceClassInstance;
-cudaMalloc((void**)&deviceClassInstance, sizeof(Derived));
-cudaMemcpy(deviceClassInstance, hostClassInstance, sizeof(Derived), cudaMemcpyHostToDevice);
+Base* hostInstance = new Derived();
+Base* deviceInstance;
+cudaMalloc((void**)&deviceInstance, sizeof(Derived));
+cudaMemcpy(deviceInstance, hostInstance, sizeof(Derived), cudaMemcpyHostToDevice);
```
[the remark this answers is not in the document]
pzehner: More about this in the comment after.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
pzehner: I think it's a good opportunity to start using it more.

pzehner:
According to @masterleinad, using virtual functions on device is not possible with SYCL.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,7 +1,16 @@
...
+```{warning}
+Using virtual functions in parallel regions is not a good idea in general. It often degrades performance, it requires specific code for a correct execution on GPU, and it is not portable on every backend. We recommend to use a different approach whenever possible.
+```
+
 Due to oddities of GPU programming, the use of virtual functions in Kokkos parallel regions can be complicated. This document describes the problems you're likely to face, where they come from, and how to work around them.
 
+Please note that virtual functions can be executed on device for the following backends:
+
+- Cuda; and
+- HIP.
```
JBludau, line 12:
```suggestion
- Cuda
- HIP
```

pzehner:
I profoundly modified the documentation page to remove mixed Cuda/Kokkos code and start with serial code.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
pzehner: Done in the latest commits.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
+
+- Cuda; and
+- HIP (with a limitation, as explained at the end).
+
+Especially, SYCL 2020 [cannot handle virtual functions](https://github.com/AlexeySachkov/llvm/blob/private/asachkov/virtual-functions-extension-spec/sycl/doc/design/VirtualFunctions.md?rgh-link-date=2024-07-16T15%3A15%3A11Z).
+
 ## The Problem
 
-In GPU programming, you might have run into the bug of calling a host function from the device. A similar thing can happen for subtle reasons in code using virtual functions. Consider the following code
+In GPU programming, you might have run into the bug of calling a host function from the device. A similar thing can happen for subtle reasons in code using virtual functions. Consider the following serial code to accelerate
```
masterleinad, line 18:
```suggestion
In GPU programming, you might have run into the bug of calling a host function from the device. A similar thing can happen for subtle reasons in code using virtual functions. Consider the following serial code
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
+  }
+
+  // cleanup
+  delete instance;
+}
 ```
 
-At a glance this should be fine, we've made a device instance of a class, copied the contents of a host instance into it, and then used it. This code will typically crash, however, because `virtualFunction` will call a host version of the function. To understand why, you'll need to understand a bit about how virtual functions are implemented.
+This code is more complex to port on GPU than it looks like.
+Using a straightforward approach, we would accelerate the `for` loop with `parallel_for` and copy `instance` on the GPU memory (not disclosing how for now).
```
masterleinad, line 49:
```suggestion
Using a straightforward approach, we would replace the `for` loop with `parallel_for` and copy `instance` on the GPU memory (not disclosing how for now).
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
+    deviceInstance->~Base(); // destroy on device
+  });
+  Kokkos::fence("wait for destroy");
+  Kokkos::kokkos_free(deviceInstanceMemory); // free
+}
 ```
 
+We first use the `KOKKOS_FUNCTION` macro to make the methods callable from a kernel.
+When creating the instance, note that we introduce a distinction between the *memory* that the it uses, and the actual instantiated *object*.
+The construct is done on the device, withing a single-iteration `parallel_for`, using placement new.
```
masterleinad, line 136:
```suggestion
The construct is done on the device, within a single-iteration `parallel_for`, using placement new.
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
+Since the kernel does not have a return type, we use a static cast to associate the object to the memory allocation.
```
masterleinad, line 137:
```suggestion
Since the kernel does not have a return type, we use a static cast to associate the object with the memory allocation.
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
+```
+
 Due to oddities of GPU programming, the use of virtual functions in Kokkos parallel regions can be complicated. This document describes the problems you're likely to face, where they come from, and how to work around them.
 
+Please note that virtual functions can be executed on device for the following backends:
+
+- Cuda; and
+- HIP (with a limitation, as explained at the end).
+
+Especially, SYCL 2020 [cannot handle virtual functions](https://github.com/AlexeySachkov/llvm/blob/private/asachkov/virtual-functions-extension-spec/sycl/doc/design/VirtualFunctions.md?rgh-link-date=2024-07-16T15%3A15%3A11Z).
```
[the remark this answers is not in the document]
pzehner: Better, indeed.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
+  }
+
+  // cleanup
+  delete instance;
+}
 ```
 
-At a glance this should be fine, we've made a device instance of a class, copied the contents of a host instance into it, and then used it. This code will typically crash, however, because `virtualFunction` will call a host version of the function. To understand why, you'll need to understand a bit about how virtual functions are implemented.
+This code is more complex to port on GPU than it looks like.
+Using a straightforward approach, we would replace the `for` loop with `parallel_for` and copy `instance` on the GPU memory (not disclosing how for now).
```
tpadioleau, line 49:
```suggestion
Using a straightforward approach, we would annotate functions with `KOKKOS_FUNCTION`, replace the `for` loop with `parallel_for` and copy `instance` on the GPU memory (not disclosing how for now).
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
+  KOKKOS_FUNCTION
+  void Bar() override {}
+};
+
+int main(int argc, char *argv[])
+{
+  Kokkos::ScopeGuard kokkos(argc, argv);
+
+  // create
+  auto deviceInstanceMemory = Kokkos::kokkos_malloc(sizeof(Derived)); // allocate memory on device
```
tpadioleau, line 112:
```suggestion
  void* deviceInstanceMemory = Kokkos::kokkos_malloc(sizeof(Derived)); // allocate memory on device
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
+  Kokkos::parallel_for("initialize", 1, KOKKOS_LAMBDA (const int i) {
+    new (static_cast<Derived*>(deviceInstanceMemory)) Derived(); // initialize on device
+  });
+  Kokkos::fence("wait for initialize");
```
tpadioleau, line 116:
```suggestion
```

There should need to fence here right ?

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
+  Base* deviceInstance = static_cast<Derived*>(deviceInstanceMemory); // declare on this memory
+
+  // use
+  Kokkos::parallel_for("myKernel", 10, KOKKOS_LAMBDA (const int i) {
+      deviceInstance->Bar();
+  });
+  Kokkos::fence();
```
tpadioleau, line 123:
```suggestion
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
+
+  // cleanup
+  Kokkos::parallel_for("destroy", 1, KOKKOS_LAMBDA (const int i) {
+    deviceInstance->~Base(); // destroy on device
+  });
+  Kokkos::fence("wait for destroy");
```
tpadioleau, line 129:
```suggestion
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
+}
 ```
 
+We first use the `KOKKOS_FUNCTION` macro to make the methods callable from a kernel.
+When creating the instance, note that we introduce a distinction between the *memory* that the it uses, and the actual instantiated *object*.
+The construct is done on the device, within a single-iteration `parallel_for`, using placement new.
+Since the kernel does not have a return type, we use a static cast to associate the object with the memory allocation.
+
+Like with other uses of `new`, we need to later free the memory.
+The destruct is done on the device, again with a single-iteration `parallel_for`.
```
tpadioleau, line 140:
```suggestion

The destructor must explicitly be called from the device, again with a single-iteration `parallel_for`. Then the memory allocation can be release with `kokkos_free`.
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
-cudaMemcpy(deviceInstance, hostInstance); // to initialize the deivce
-Kokkos::parallel_for(... {
-  // use deviceInstance
-});
-```
+## What if I need a setter that works with host values?
 
-We can't translate this easily, the naive translation would be
+The first problem people run into with this is that they want to initialize some fields based on host data, with a setter which is *not* a virtual function.
+Calling this setter on the device would crash if the host data is not easily copyable on the device (e.g. for a small array).
```
tpadioleau, line 152:
```suggestion
Calling this setter on the device would crash if the host data is not easily copyable on the device.
```

I don't understand the difficulty with small arrays ?

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,224 @@
...
+The most productive solution we've found in these cases is to allocate the instance in `SharedSpace`, initialize it on the device, and then fill in fields on the host
 
 ```c++
-Base* deviceInstance; // cudaMalloc'd to allocate it
-Kokkos::parallel_for(... {
-  new((Derived*)deviceInstance) Derived(); // initialize an instance, and place the result in the pointer deviceInstance
-  deviceInstance->setAField(someHostValue);
-});
+// create
+auto deviceInstanceMemory = Kokkos::kokkos_malloc<Kokkos::SharedSpace>(sizeof(Derived)); // allocate on shared space
```
tpadioleau, line 157:
```suggestion
void* deviceInstanceMemory = Kokkos::kokkos_malloc<Kokkos::SharedSpace>(sizeof(Derived)); // allocate on shared space
```

tpadioleau (changes requested):

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
pzehner: I was wondering if it was better to keep it `auto`, because `void*` is kind of a joker type.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
pzehner: Are you sure we don't need it?

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
pzehner: This is a remaining of the last iteration. You could do, as the text says:

```cpp
Kokkos::parallel_for("myLoop", 1, KOKKOS_LAMBDA (const int i) {
  deviceInstance.setAField(hostValue);
});

But if `hostValue` is not a scalar type (like the pointer of an array), the code will crash.

Comment (tpadioleau) [reply docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md:151]: I see, I don't feel that mentioning "small array" helps in the end

File: docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md (hunk 14 above)
Comment (tpadioleau) [reply docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md:116]: As long as you are not manipulating different execution spaces you are fine. Given your code, they execute in order in the default execution space.

File: docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md (hunk 13 above)
Comment (tpadioleau) [reply docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md:112]: Well it is the return type of `kokkos_malloc`

File: docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md (hunk 18 above)
Comment (JBludau) [reply docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md:151]: there could be multiple things wrong here. 

- The lambda capture could result in an invalid instance on the device.
- The type itself could not be copyable or assignable on the device.
- Side effects could happen in at unexpected times, because async kernel run.

So maybe we could write:
```suggestion
Calling this setter on the device would crash if `hostValue` does not form a valid type on the device 
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
pzehner: Ok, fair.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
pzehner: Are you sure for this one?

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
pzehner: I think it would be more pedagogical to explicit what could go wrong. When I read the page before this PR, I didn't get why the capture of the lambda wouldn't be enough.

pzehner: I tried another formulation based on @JBludau's remark.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
cedricchevalier19: For this one, the `kokkos_free()` documentation is unclear. Perhaps it should be added to another PR whether it requires fencing or not: it is a host function that affects device memory. It looks like Kokkos has some internal fences for that, but Cuda API does not.

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -147,8 +147,8 @@ For a full working example, see [the example in the repo](https://github.com/kok
 
 ## What if I need a setter that works with host values?
 
-The first problem people run into with this is that they want to initialize some fields based on host data, with a setter which is *not* a virtual function.
-Calling this setter on the device would crash if the host data is not easily copyable on the device (e.g. for a small array).
+The first problem people run into with this is when they want to initialize some fields based on host data, using a setter which is *not* a virtual function.
+Calling this setter on the device would crash if the host data doesn't form a valid type on the device: especially, if the data type is not copyable on the device (e.g. for a pointer), or if the data instance is invalid on the device.
```
JBludau, line 151: I would prefer to leave the part post :. I think it is just confusing or a repetition. 
A pointer is a valid type on device and if it points into the right memory space, the setter is perfectly valid. Even assigning a pointer to host memory on the device is valid as the violation will only occur if the pointer is read/written to.

If you want to specify anything here, I would suggest spelling out the abstract requirements. If you want an example, you could refer to the requirements and explain what is violated

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,231 @@
...
 
-Base* hostClassInstance = new Derived();
-Base* deviceClassInstance;
-cudaMalloc((void**)&deviceClassInstance, sizeof(Derived));
-cudaMemcpy(deviceClassInstance, hostClassInstance, sizeof(Derived), cudaMemcpyHostToDevice);
 
-Kokkos::parallel_for("DeviceKernel", SomeCudaPolicy, KOKKOS_LAMBDA(const int i) {
-  deviceClassInstance->Bar();
-});
+int main(int argc, char *argv[])
```
JBludau, line 34:
```suggestion
int main(int argc, char *argv[])
{
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,231 @@
...
 ![VPointer](./figures/VirtualFunctions-VPointers.png)
 
-Now that we know what the compiler is doing to implement virtual functions, we'll look at why it doesn't work with GPU's
+Now that we know what the compiler is doing to implement virtual functions, we'll look at why it doesn't work with GPU's.
 
-Credit: the content of this section is adapted from Pablo Arias [here](https://pabloariasal.github.io/2017/06/10/understanding-virtual-tables/)
+Credit: the content of this section is adapted from Pablo Arias [here](https://pabloariasal.github.io/2017/06/10/understanding-virtual-tables/).
 
-## Then why doesn't my code work?
+## Then why the straightforward approach doesn't work?
```
JBludau, line 70:
```suggestion
## Then why doesn't the straightforward approach work?
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,231 @@
...
 ![VPointerToHost](./figures/VirtualFunctions-VPointerToHost.png)
 
-Our cudaMemcpy faithfully copied all of the members of the class, including the Vpointer merrily pointing at host functions, which we then call on the device. 
+We faithfully copied all of the members of the class on the GPU memory, including the Vpointer happily pointing at host functions, which we then call on the device.
 
-## How to fix this
+## Make it work
 
-The problem here is that we are constructing the class on the Host. If we were constructing on the Device, we'd get the correct Vpointer, and thus the correct functions (but only for calls on the device). In pseudocode, we want to move from
+The problem here is that we are constructing the instance on the Host.
```
JBludau, line 84:
```suggestion
The problem here is that we are constructing the instance on the host.
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,231 @@
...
+If we were constructing it on the Device, we'd get the correct Vpointer, and thus the correct functions.
```
JBludau, line 85:
```suggestion
If we were constructing it on the device, we'd get the correct Vpointer, and thus the correct functions.
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,231 @@
...
 
-```c++
-Base* hostInstance = new Derived(); // allocate and initialize host
-Base* deviceInstance; // cudaMalloc'd to allocate
-cudaMemcpy(deviceInstance, hostInstance); // to initialize the deivce
-Kokkos::parallel_for(... {
-  // use deviceInstance
-});
-```
+To that aim, we first allocate memory on the device, then construct on the device using a technique called *placement new*
```
JBludau, line 88:
```suggestion
Therefore, we first allocate memory on the device, then construct on the device using a technique called `placement new <https://en.cppreference.com/w/cpp/language/new#Placement_new> _`.
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,231 @@
...
+  virtual void Bar() {}
+};
+
 class Derived : public Base {
-  /** fields */
   public:
-  KOKKOS_FUNCTION virtual void Bar(){
-    // TODO: implement all of physics
-  } 
+  void Bar() override {}
```
JBludau, line 30:
```suggestion
  public:
  void Bar() override {}
```
for consistency with later code on this page

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,231 @@
...
+  Kokkos::kokkos_free(deviceInstanceMemory); // free
+
+  }
+  Kokkos::finalize();
+}
 ```
 
+We first use the `KOKKOS_FUNCTION` macro to make the methods callable from a kernel.
+When creating the instance, note that we introduce a distinction between the *memory* that the it uses, and the actual instantiated *object*.
+The construct is done on the device, within a single-iteration `parallel_for`, using placement new.
```
JBludau, line 138:
```suggestion
The construction is done on the device, within a single-iteration `parallel_for`, using `placement new <https://en.cppreference.com/w/cpp/language/new#Placement_new> _`.
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,231 @@
...
+Since the kernel does not have a return type, we use a static cast to associate the object with the memory allocation.
```
JBludau, line 139:
```suggestion
Since the kernel does not have a return type, we use a static cast to associate the object type with the memory allocation.
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,231 @@
...
+
+The destructor must explicitly be called from the device, again with a single-iteration `parallel_for`. Then the memory allocation can be release with `kokkos_free`.
```
JBludau, line 141:
```suggestion
For not `trivially destructable https://en.cppreference.com/w/cpp/language/destructor#Trivial_destructor`_ objects the destructor must explicitly be called on the device. After destructing the object in a single-iteration `parallel_for` the memory can be release with `kokkos_free`.
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,231 @@
...
 
-Note that like with other uses of `new`, you need to later `free` the memory.
 For a full working example, see [the example in the repo](https://github.com/kokkos/kokkos/blob/master/example/virtual_functions/main.cpp).
 
-## Complications and Fixes
+## What if I need a setter that works with host values?
 
-The first problem people run into with this is that they want to initialize some fields or nested classes based on host data before moving data down to the device
+The first problem people run into with this is when they want to initialize some fields based on host data, using a setter which is *not* a virtual function.
+Calling this setter on the device would crash if the host data doesn't form a valid type on the device: especially, if the data type is not copyable on the device (e.g. for a pointer), or if the data instance is invalid on the device.
```
JBludau, line 153:
```suggestion
The first problem people run into with this is when they want to set some fields based on host data. As the object instance resides in device memory, it might not be accessible by the host. But the fields can be set within a `parallel_for` on the device. Nevertheless, this requires that the lambda or functor that sets the fields on the device must have access to the host data.
```

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,184 +1,231 @@
...
+The most productive solution we've found in these cases is to allocate the instance in `SharedSpace`, initialize it on the device, and then fill in fields on the host
```
JBludau, line 154:
```suggestion
The most productive solution we've found in these cases is to allocate the object instance in `SharedSpace`, which allows to construct it on the device, and then set fields on the host
```

JBludau (commented):
I only have minor nitpicking

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
cedricchevalier19: I am skeptical about this one. `Construct` is not very clear to me.

JBludau: I was aiming on: construct = call the constructor

docs/source/ProgrammingGuide/Kokkos-and-Virtual-Functions.md
```diff
@@ -1,186 +1,233 @@
...
+  Kokkos::finalize();
+}
 ```
 
+We first use the `KOKKOS_FUNCTION` macro to make the methods callable from a kernel.
+When creating the instance, note that we introduce a distinction between the *memory* that it uses, and the actual instantiated *object*.
+The object instance is constructed on the device, within a single-iteration `parallel_for`, using [placement new](https://en.cppreference.com/w/cpp/language/new#Placement_new).
+Since the kernel does not have a return type, we use a static cast to associate the object type with the memory allocation.
+
+For not [trivially destructable](https://en.cppreference.com/w/cpp/language/destructor#Trivial_destructor) objects the destructor must explicitly be called on the device.
```
tpadioleau, line 141:
```suggestion
For not [trivially destructible](https://en.cppreference.com/w/cpp/language/destructor#Trivial_destructor) objects the destructor must explicitly be called on the device.
```

tpadioleau (approved):
I find the page quite clear!

JBludau (approved):
Looks good to me

dalg24:
Please squash merge next time

PR #552, opened 2024-07-12
List backend-specific `IndexType` for `ExecutionPolicy`

JBludau:
In a discussion on slack it came up, that the `IndexType` of `ExecutionPolicy` does not always default to `int64_t` as stated in the documentation.

From looking at the implementation it defaults to the `size_type` of the `MemorySpace` that is associated with the `ExecutionSpace` given to the `ExecutionPolicy`. Therefore, this PR adds a table with the backend-specific `size_type` to the `MemorySpace` page and adds a link to it in the common arguments page of the `ExecutionPolicy` page.

### DIFF
--- a/docs/source/API/core/Execution-Policies.rst
+++ b/docs/source/API/core/Execution-Policies.rst
@@ -82,7 +82,7 @@ Execution Policies generally accept compile time arguments via template paramete
 
     * * IndexType
       * ``IndexType<int>``
-      * Specify integer type to be used for traversing the iteration space. Defaults to ``int64_t``.
+      * Specify integer type to be used for traversing the iteration space. Defaults to the ``size_type`` of `MemorySpaceConcept <memory_spaces.html#typedefs>`__.
 
     * * LaunchBounds
       * ``LaunchBounds<MaxThreads, MinBlocks>``
--- a/docs/source/API/core/Execution-Policies.rst
+++ b/docs/source/API/core/Execution-Policies.rst
@@ -81,8 +81,8 @@ Execution Policies generally accept compile time arguments via template paramete
       * Specify scheduling policy for work items. ``Dynamic`` scheduling is implemented through a work stealing queue. Default is machine and backend specific.
 
     * * IndexType
-      * ``IndexType<int>``
-      * Specify integer type to be used for traversing the iteration space. Defaults to the ``size_type`` of `ExecutionSpaceConcept <execution_spaces.html#typedefs>`__.
+      * e.g. ``IndexType<int>``
+      * Specify integer type to be used for traversing the iteration space. Defaults to the ``size_type`` of `ExecutionSpaceConcept <execution_spaces.html#typedefs>`__. Can affect the performance depending on the backend.
 
     * * LaunchBounds
       * ``LaunchBounds<MaxThreads, MinBlocks>``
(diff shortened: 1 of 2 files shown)


docs/source/API/core/memory_spaces.rst
```diff
@@ -152,6 +153,26 @@ Typedefs
...
+ :header-rows: 1
+
+ * - Space
+   - ``size_type``
+ * - ``Cuda*``
+   - ``unsigned int``
+ * - ``HIP*``
+   - ``unsigned int``
+ * - ``SYCL*``
+   - ``int``
```
masterleinad, line 169:
```suggestion
 * - ``Cuda``
   - ``unsigned int``
 * - ``HIP``
   - ``unsigned int``
 * - ``SYCL``
   - ``int``
```
or am I missing why you are adding `*` to these three backends?

JBludau: hmm it was due to the type being linked to the `MemorySpace`, thus I added the wildcard.
But I agree, that it feels like it should be backend-specific (so far it is, but each `MemorySpace` impl could define it differently). We could decide to just communicate it as backend-specific, which we fulfill at the moment.

docs/source/API/core/Execution-Policies.rst
```diff
@@ -82,7 +82,7 @@ Execution Policies generally accept compile time arguments via template paramete
 
     * * IndexType
       * ``IndexType<int>``
-      * Specify integer type to be used for traversing the iteration space. Defaults to ``int64_t``.
+      * Specify integer type to be used for traversing the iteration space. Defaults to the ``size_type`` of `MemorySpaceConcept <memory_spaces.html#typedefs>`__.
```
masterleinad, line 85:
```suggestion
      * Specify integer type to be used for traversing the iteration space. Defaults to the ``size_type`` of the execution space.
```

docs/source/API/core/memory_spaces.rst
masterleinad: It's not quite clear to me why a memory space needs to have a `size_type` (as opposed to the execution space)` and I would rather list that member there or at least additionally.

JBludau: changed to point to the execution space now

docs/source/API/core/execution_spaces.rst
```diff
@@ -310,6 +311,27 @@ Typedefs
...
+ * - ``HIP``
+   - ``unsigned int``
+ * - ``SYCL``
+   - ``int``
+ * - ``OpenACC``
+   - ``size_t``
+ * - ``OpenMPTarget``
+   - ``unsigned int``
+ * - All host spaces
+   - ``size_t``
```
dalg24, line 333: I am not sure whether I would want to spell out what these currently are.
Maybe instead we could comment that they are implementation defined and that backends effectively have picked different signed or unsigned integer types with width 32 or 64 bits.

masterleinad: I agree.

dalg24: Sounds good

masterleinad (approved):

docs/source/API/core/execution_spaces.rst
JBludau: done

masterleinad (approved):

ajpowelsnl (commented):
Looks good!  All done?  Do the links work?

JBludau:
> Looks good! All done? Do the links work?

yeah, all done

PR #553, opened 2024-07-12
favor link to common template arguments page over copy of the table in many `ExecutionPolicies`

JBludau:
It just makes more sense to not repeat ourselves I think...
Came up in the wake of #552


masterleinad (approved):

ajpowelsnl:
Anything else needed for this PR ?  Have we checked to make sure the links work : )

PR #558, opened 2024-07-30
Add a note about the indices conversion checks in RangePolicy

ldh4:
Changed argument index types from `int64_t` to `IndexType` to better describe what's really in the implementation, which is templated type that is required to be convertible to `IndexType`.
It's already noted that the internal `IndexType` will be defaulted to `int64_t` if not specified.

Also added a note about changed behavior in conversion safety check when roundtrip convertibility is not provided (from https://github.com/kokkos/kokkos/pull/7172).

### DIFF
--- a/docs/source/API/core/policies/RangePolicy.rst
+++ b/docs/source/API/core/policies/RangePolicy.rst
@@ -120,19 +120,19 @@ Constructors
 
    Default Constructor uninitialized policy.
 
-.. cppkokkos:function:: RangePolicy(int64_t begin, int64_t end)
+.. cppkokkos:function:: RangePolicy(IndexType begin, IndexType end)
 
    Provide a start and end index.
 
-.. cppkokkos:function:: RangePolicy(int64_t begin, int64_t end, ChunkSize chunk_size)
+.. cppkokkos:function:: RangePolicy(IndexType begin, IndexType end, ChunkSize chunk_size)
 
    Provide a start and end index as well as a ``ChunkSize``.
 
-.. cppkokkos:function:: RangePolicy(const ExecutionSpace& space, int64_t begin, int64_t end)
+.. cppkokkos:function:: RangePolicy(const ExecutionSpace& space, IndexType begin, IndexType end)
 
    Provide a start and end index and an ``ExecutionSpace`` instance to use as the execution resource.
 
-.. cppkokkos:function:: RangePolicy(const ExecutionSpace& space, int64_t begin, int64_t end, ChunkSize chunk_size)
+.. cppkokkos:function:: RangePolicy(const ExecutionSpace& space, IndexType begin, IndexType end, ChunkSize chunk_size)
 
    Provide a start and end index and an ``ExecutionSpace`` instance to use as the execution resource, as well as a ``ChunkSize``.
 
@@ -141,6 +141,7 @@ Preconditions:
 
 * The start index must not be greater than the end index.
 * The actual constructors are templated so we can check that they are converted to ``index_type`` safely (see `#6754 <https://github.com/kokkos/kokkos/pull/6754>`_).
+   * The conversion safety check is only performed if ``index_type`` is convertible to the start and end index types.
 
 CTAD Constructors (since 4.3):
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


docs/source/API/core/policies/RangePolicy.rst
```diff
@@ -141,6 +141,7 @@ Preconditions:
 
 * The start index must not be greater than the end index.
 * The actual constructors are templated so we can check that they are converted to ``index_type`` safely (see `#6754 <https://github.com/kokkos/kokkos/pull/6754>`_).
+   * The conversion safety check is only performed if ``index_type`` is convertible to the start and end index types.
```
dalg24, line 144: start -> begin for consistency

ldh4: Looks like there were already mixed usages of start/end and begin/end throughout this document.
Changed all `start` to `begin`.  6cd5e99

dalg24:
> Changed argument index types from `int64_t` to `IndexType` to better describe what's really in the implementation, which is templated type that is required to be convertible to `IndexType`.

Should we not just it is templated?  (not asking you to change, just trying to follow the reasoning)

ldh4:
> > Changed argument index types from `int64_t` to `IndexType` to better describe what's really in the implementation, which is templated type that is required to be convertible to `IndexType`.
> 
> Should we not just it is templated? (not asking you to change, just trying to follow the reasoning)

I am for exposing that `IndexType` is a templated type. It's just more accurate that way, but I think if we are to do that, few corrections also should be made throughout this documentation page. More specifically, the fact that `index_type` and `IndexType` are somewhat used interchangeably throughout this page should be fixed (including in synopsis).

Constructors should either be
```
template <typename IndexType1, typename IndexType2,
          std::enable_if_t<(std::is_convertible_v<IndexType1, index_type> &&
          std::is_convertible_v<IndexType2, index_type>),
          bool> = false>
RangePolicy(IndexType1 work_begin, IndexType2 work_end );
```
or maybe just:
```
template <typename IndexType1, typename IndexType2>
RangePolicy(IndexType1 work_begin, IndexType2 work_end );
```
If either option sounds reasonable, I will go ahead and add these changes.

nliber:
I don't know the right way to do this.

The templated constructors are really just an implementation detail.  It is designed to work like non-templated overloads in that we are simulating implicit convertibility to `member_type` with the additional benefit that we check that no information is being lost.  I'd rather we figure out some way to say that.  I don't want to say `member_type` as that is just a bit too abstract, as most people are just using the defaults, passing numbers, and not thinking about it in generic terms.

nliber:
Expanding on what I wrote on Slack:

We need a different name than `IndexType`, as that is also the class template name one passes to change the internal type.  We can change it to either `index_type` or `member_type`.

Maybe add a note saying it defaults to the `size_type` of the `execution_space`, which is typically an unsigned type, and can be changed by passing in `IndexType<T>` as a template parameter to `RangePolicy`.  I know this is redundant with the description at [Execution Policies](https://kokkos.org/kokkos-core-wiki/API/core/Execution-Policies.html), but there is a lot to read on there to figure this out.

Even though they are interchangeable, we really should be consistent on which one we use in the docs.  I have a slight preference for `index_type` (the downside being that `IndexType` and `index_type` are pronounced the same way, even in my head).

We need to balance simplifying the documentation for users w/o glossing over lower level details they may eventually need.

ldh4:
@nliber 

> Expanding on what I wrote on Slack:
> 
> We need a different name than `IndexType`, as that is also the class template name one passes to change the internal type. We can change it to either `index_type` or `member_type`.

I overlooked that there's `Kokkos::IndexType`. Reverted to `index_type`.

> Maybe add a note saying it defaults to the `size_type` of the `execution_space`, which is typically an unsigned type, and can be changed by passing in `IndexType<T>` as a template parameter to `RangePolicy`. I know this is redundant with the description at [Execution Policies](https://kokkos.org/kokkos-core-wiki/API/core/Execution-Policies.html), but there is a lot to read on there to figure this out.

This information is already available on this page in `Common Arguments for all Execution Policies` section.
Modified `IndexType` row to make its description consistent with what's described in `Execution Policies` page.

> Even though they are interchangeable, we really should be consistent on which one we use in the docs. I have a slight preference for `index_type` (the downside being that `IndexType` and `index_type` are pronounced the same way, even in my head).
> 
> We need to balance simplifying the documentation for users w/o glossing over lower level details they may eventually need.

Although I do also prefer `index_type` over `member_type`, `index_type` is not used anymore in our implementation. Only `member_type` is used.
It doesn't really make any difference since `member_type` and `index_type` are the same things, but it makes me wonder if we should prefer `member_type` everywhere or refactor RangePolicy to use `index_type` over `member_type` to avoid confusion.
One problem I have with `member_type` is that its naming doesn't directly give an idea on what member it is referring to.

nliber:
If a reader asks "what is `member_type` or `index_type`?"

If you look at the synopsis, you see
```c++
using member_type = PolicyTraits<Args...>::index_type;
//...
// Inherited from PolicyTraits<Args...>
//...
using index_type = PolicyTraits<Args...>::index_type;
```

It's a clue, but not particularly helpful.  You have to dig through the source to understand what `PolicyTraits` does (good luck).  More likely, if we use `index_type`, at least people can guess correctly that they may be related.  But you still have to guess that it is either `IndexType<T>` if specified, or "`size_type` of `ExecutionSpaceConcept`" if not (which they then have to guess maps to `execution_space::size_type`).  If it is `execution_space::size_type`, it is probably the unsigned `size_t` (although in reality it is sometimes it is `unsigned`, which may or may not match `size_t`), because that is what it tends to be in the C++ standard library.

That is a lot of mental gymnastics to go through to reason out that "I need numbers and they should probably be unsigned" (and even the "unsigned" part people skip; otherwise, we wouldn't need the round trip test).

Brainstorming here:  maybe we just need to say something like:

- `index_type` - If `IndexType<T>` is passed as a template parameter to `RangePolicy` then `T`; otherwise `execution_space::size_type` (which is usually an unsigned number type)
- `member_type` - a synonym for `index_type`

docs/source/API/core/policies/RangePolicy.rst
```diff
@@ -120,27 +120,28 @@ Constructors
...
-   Provide a start and end index and an ``ExecutionSpace`` instance to use as the execution resource, as well as a ``ChunkSize``.
+   Provide a begin and end index and an ``ExecutionSpace`` instance to use as the execution resource, as well as a ``ChunkSize``.
 
 Preconditions:
 ^^^^^^^^^^^^^^
 
-* The start index must not be greater than the end index.
+* The begin index must not be greater than the end index.
 * The actual constructors are templated so we can check that they are converted to ``index_type`` safely (see `#6754 <https://github.com/kokkos/kokkos/pull/6754>`_).
+   * The conversion safety check is performed only if ``index_type`` is convertible to the begin and end index types.
```
nliber, line 144:
```suggestion
* The actual constructors are templated so we can check that they are converted to ``index_type`` safely if possible (see `#6754 <https://github.com/kokkos/kokkos/pull/6754>`_).
```

Line 144 doesn't work, for the following reasons:

- You have to mention the template types if you want to describe the mechanism of converting to `index_type`.  Saying that `index_type` is convertible to the begin index type, which is `index_type`, doesn't make sense (or course a type is convertible to itself, but that isn't what you are trying to say).
- The conversion check cannot be performed for a type that isn't a fundamental type or a type Kokkos knows about.  For a UDT, even if you can construct it with an `index_type`, you don't know that it has the same value it started with.

I think it is better to just say we make a best effort at checking the conversion.

nliber (changes requested):

docs/source/API/core/policies/RangePolicy.rst
```diff
@@ -120,27 +120,28 @@ Constructors
 
    Default Constructor uninitialized policy.
 
-.. cppkokkos:function:: RangePolicy(int64_t begin, int64_t end)
+.. cppkokkos:function:: RangePolicy(index_type begin, index_type end)
```
nmm0, line 123: We need to add a `versionchanged` directive: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionchanged

ldh4: Thanks @nmm0 ! I was looking for a directive for noting a version just like that one.
But as for this line, it's a bit misleading since the argument types for this constructor were never really `int64_t`. They were templated `member_type` before which were defaulted to `int64_t`, if the policy wasn't constructed with any specified index type. So, the arg type hasn't changed, and this might be more of a correction in the docs.

docs/source/API/core/policies/RangePolicy.rst
ldh4: I agree that it might just be better for us to state that we try to make best effort on this. There weren't really good ways to summarize the general behavior, especially if UDTs are involved, in which case the conversion checks had to be completely skipped.

docs/source/API/core/policies/RangePolicy.rst
nmm0: Ok makes sense, I think it's not needed then

nliber (approved):

PR #561, opened 2024-08-06
Add note about ChunkSize explicit constructor since 4.4

tcclevenger:
I chose to use a "note", but I'm happy to switch to a "warning" or bold text (done other places) or other suggestions.

### DIFF
--- a/docs/source/API/core/policies/RangePolicy.rst
+++ b/docs/source/API/core/policies/RangePolicy.rst
@@ -30,7 +30,7 @@ Synopsis
 .. code-block:: cpp
 
     struct Kokkos::ChunkSize {
-        ChunkSize(int value_);
+        explicit ChunkSize(int value_);
     };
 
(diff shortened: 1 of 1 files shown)


nliber (approved):
I think that is fine.  While the code fix is a breaking change, my expectation is that virtually no one took advantage of the previous behavior and if they did, the change to their code is minimal and not unexpected when working with C++.

dalg24 (commented):
We don't want the `explicit` specifier here https://github.com/kokkos/kokkos-core-wiki/pull/561/files#diff-1e20fae80dae2837b5fdb19d79a210e4c1602b9e060f53f90f315eaa794a18d1R98
?

**edit** I mean besides the note you added

tcclevenger:
@dalg24 added `explicit` specifier.

masterleinad (approved):

PR #567, opened 2024-08-26
Add missing documentation for UnorderedMap

tcclevenger:
Documentation was missing for the following public functions
- `create_copy_view(UnorderedMap src)`
- `allocate_view(UnorderedMap src)`
- `deep_copy_view(UnorderedMap src)`

and non-member functions
- `deep_copy(UnorderedMap dst, UnorderedMap src)`
- `UnorderedMap::HostMirror create_mirror(UnorderedMap src)`

### DIFF
--- a/docs/source/API/containers/Unordered-Map.rst
+++ b/docs/source/API/containers/Unordered-Map.rst
@@ -96,6 +96,29 @@ Description
 
       Return true if the internal views (keys, values, hashmap) are allocated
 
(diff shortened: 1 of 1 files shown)


crtrott (approved):
Oh my I almost don't want to approve because I think it's a mess that we got these member functions. But since we had them forever we should document and deprecate.

tcclevenger:
> Oh my I almost don't want to approve because I think it's a mess that we got these member functions. But since we had them forever we should document and deprecate.

I'll put it on my todo list.

PR #569, opened 2024-08-27
core(view): warn about the implications of without initializing

romintomasetti:
This PR describes what `Kokkos::WithoutInitialization` implies.

### DIFF
--- a/docs/source/ProgrammingGuide/View.rst
+++ b/docs/source/ProgrammingGuide/View.rst
@@ -97,7 +97,9 @@ Finally, note that virtual functions are technically allowed, but calling them i
 6.2.3 Can I make a View of Views?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-NOVICES: THE ANSWER FOR YOU IS "NO."  PLEASE SKIP THIS SECTION.  
+.. warning::
+
+  NOVICES: THE ANSWER FOR YOU IS "NO."  PLEASE SKIP THIS SECTION.  
 
 A "View of Views" is a special case of View, where the type of each entry is itself a View. It is possible to make this, but before you try, please see below.
 
--- a/docs/source/ProgrammingGuide/View.rst
+++ b/docs/source/ProgrammingGuide/View.rst
@@ -460,6 +460,14 @@ in a way that reflects how first-touch affinity to threads is desired.
 Typically it is sufficient to use the parallel iteration index as the index of the location in the
 view to write to.
 
+.. warning::
+
+  :cppkokkos:`WithoutInitialization` implies that the destructor of each element of the :cppkokkos:`View` **will not be called**.
+  For instance, if the :cppkokkos:`View`'s value type is not trivially destructible,
+  you **should not use** :cppkokkos:`WithoutInitialization` unless you are taking care of calling the destructor manually before the :cppkokkos:`View` deallocates its memory.
+
+  The mental model is that whenever placement new is used to call the constructor, the destructor also isn't called before the memory is deallocated but it needs to be called manually.
+
 6.4.3 Deep copy and HostMirror
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


docs/Makefile
```diff
@@ -1,5 +1,5 @@
 html:
-	sphinx-build -b html ./source/ ./generated_docs/ -W --keep-going
+	sphinx-build -W -b html ./source/ ./generated_docs/ -W --keep-going
```
romintomasetti, line 2: Probably good to extract this in a separate PR.

docs/source/ProgrammingGuide/View.rst
```diff
@@ -0,0 +1,723 @@
+.. role:: cppcode(code)
```
romintomasetti, line 1: I appears like I've deleted the file and re-created it, but I think it's because there are so many changes that `git` thinks that's what I've done.

dalg24 (dismissed):
Please do not mix the change of format with editing the actual content in the same PR

masterleinad (changes requested):
Please split this pull request up. Either add the changes to initialization first or convert the file first. Also adding another flag in `doc/Makefile` should be a separate pull request.

docs/Makefile
romintomasetti: See #570.

docs/source/ProgrammingGuide/View.rst
romintomasetti: Outdated.

docs/source/ProgrammingGuide/View.rst
```diff
@@ -458,6 +460,15 @@ in a way that reflects how first-touch affinity to threads is desired.
 Typically it is sufficient to use the parallel iteration index as the index of the location in the
 view to write to.
 
+.. warning::
+
+  :cppkokkos:`WithoutInitialization` implies that the destructor of each element of the :cppkokkos:`View` **will not be called**.
+  For instance, if you need the destructor of your :code:`MyObjectWithDestructor` be called for each element of your :cppkokkos:`View<MyObjectWithDestructor*>`,
+  you **should not use** :cppkokkos:`WithoutInitialization`.
```
masterleinad, line 467:
```suggestion
  For instance, if the View's value type is not trivially destructible,
  you **should not use** :cppkokkos:`WithoutInitialization` unless you are taking care of calling the destructor manually before the View deallocates its memory.
```

docs/source/ProgrammingGuide/View.rst
```diff
@@ -458,6 +460,15 @@ in a way that reflects how first-touch affinity to threads is desired.
...
+
+  The reason for that is, since `Kokkos` did not initialize the memory, it cannot know onto which elements of the :cppkokkos:`View` calling the destructor
+  of the :code:`value_type` is valid.
```
masterleinad, line 470: Maybe, rather something like
```suggestion
The mental model is that whenever placement new is used to call the constructor, the destructor also isn't called before the memory is deallocated but it needs to be be called manually.
```
Or omit it with my previous suggestion.

docs/source/ProgrammingGuide/View.rst
romintomasetti: Much better indeed.

docs/source/ProgrammingGuide/View.rst
romintomasetti: :heavy_check_mark:

masterleinad (approved):

docs/source/ProgrammingGuide/View.rst
```diff
@@ -458,6 +460,14 @@ in a way that reflects how first-touch affinity to threads is desired.
 Typically it is sufficient to use the parallel iteration index as the index of the location in the
 view to write to.
 
+.. warning::
+
+  :cppkokkos:`WithoutInitialization` implies that the destructor of each element of the :cppkokkos:`View` **will not be called**.
+  For instance, if the :cppkokkos:`View`'s value type is not trivially destructible,
+  you **should not use** :cppkokkos:`WithoutInitialization` unless you are taking care of calling the destructor manually before the :cppkokkos:`View` deallocates its memory.
+
+  The mental model is that whenever placement new is used to call the constructor, the destructor also isn't called before the memory is deallocated but it needs to be called manually.
```
ajpowelsnl, line 469: **_Suggestion_**:  maybe spell out what to do in the case of both trivially and non-trivially destructible data types, maybe even including a couple of simple code-snippet examples.  Perhaps also mention a couple of typical use cases where someone would want to use a View `WithoutInitialization`.

The sentence on line 469 is not clear, and is a little redundant with the immediately preceding statements.  It may not be necessary if the previous verbiage were sharpened up, with spelled out use cases and an couple of simple examples.

ajpowelsnl (commented):
Good instinct to improve this documentation.  Cleaning up the `WithoutInitialization` section is the more actionable suggestion.  My comment about "novice" is one for broader consideration.

docs/source/ProgrammingGuide/View.rst
```diff
@@ -97,7 +97,9 @@ Finally, note that virtual functions are technically allowed, but calling them i
 6.2.3 Can I make a View of Views?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-NOVICES: THE ANSWER FOR YOU IS "NO."  PLEASE SKIP THIS SECTION.  
+.. warning::
+
+  NOVICES: THE ANSWER FOR YOU IS "NO."  PLEASE SKIP THIS SECTION.
```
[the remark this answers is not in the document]
dalg24: This is beyond the scope of this PR.  The text is not being changed, just the typeset.

crtrott (approved):
I think this is sufficient. I you don't understand the difference between trivially destructible and not trivially destructible you should not use View of Views - i.e. you are a `novice` and should have skipped this section.

PR #571, opened 2024-08-28
core(view): changes due to conversion from md to rst

romintomasetti:
As mentioned in #569, I'll need the `View.md` to understand `RST`. So I need to renamed it. And so `Sphinx` complains.

Changes:
1. Sections, subsections and so on.
2. Code blocks.
3. References (to sections, footnote).
4. Italic.
5. The table becomes a `csv-table`.

### DIFF
--- a/docs/source/ProgrammingGuide/View.rst
+++ b/docs/source/ProgrammingGuide/View.rst
@@ -1,4 +1,5 @@
-# 6: View: Multidimensional array
+6: View: Multidimensional array
+===============================
 
 After reading this chapter, you should understand the following:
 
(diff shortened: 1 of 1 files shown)


romintomasetti:
@dalg24 @masterleinad The diff is still hard to read, even though I went over the `Sphinx` complains one by one (nothing more). What should we do ?

masterleinad (approved):
Looks good to me. Ignoring white space changes makes this not that hard to compare.

dalg24:
Did you check that the output looks reasonable?

masterleinad:
The documentation in the artifact looked fine to me. We could try to improve the accessibility matrix a bit and `reference_is_lvalue` should be `reference_type_is_lvalue` but we can address that in a follow-up since it's preexisting.

PR #574, opened 2024-08-29
Document version macros

dalg24:


dalg24:
(Fixed typo `if (KOKKOS_VERSION_PATCH = 99)` -> `==`)

docs/source/API/core/Macros.rst
```diff
@@ -4,11 +4,71 @@ Macros
...
-| ``KOKKOS_VERSION`` | The Kokkos version; ``KOKKOS_VERSION % 100`` is the patch level, ``KOKKOS_VERSION / 100 % 100`` is the minor version, and ``KOKKOS_VERSION / 10000`` is the major version.   |
-+--------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+.. list-table::
+   :header-rows: 1
+
+   * - Version integer macros
+   * - ``KOKKOS_VERSION``
+   * - ``KOKKOS_VERSION_MAJOR``
+   * - ``KOKKOS_VERSION_MINOR``
+   * - ``KOKKOS_VERSION_PATCH``
```
masterleinad, line 14: I don't like the looks of the one column tables very much and would either replace them with a list or add a second column that demonstrates their definition for some release.

docs/source/API/core/Macros.rst
```diff
@@ -4,11 +4,78 @@ Macros
...
-| ``KOKKOS_VERSION`` | The Kokkos version; ``KOKKOS_VERSION % 100`` is the patch level, ``KOKKOS_VERSION / 100 % 100`` is the minor version, and ``KOKKOS_VERSION / 10000`` is the major version.   |
-+--------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+.. list-table::
+   :header-rows: 1
+   :align: left
+
+   * - Version integer macros
+     - Example value assuming Kokkos version 4.2.1
+   * - ``KOKKOS_VERSION``
+     - ``4'02'01``
```
masterleinad, line 14:
```suggestion
     - ``40201``
```

dalg24: Why?

masterleinad: That's the literal value, isn't it?

dalg24: From https://en.cppreference.com/w/cpp/language/integer_literal

> Optional single quotes (`'`) may be inserted between the digits as a separator; they are ignored when determining the value of the literal.

masterleinad: I now that but I think it's more confusing for the average user looking at this page.

dalg24: I dropped the separators

masterleinad (approved):

PR #556, opened 2024-07-26
Add example code referenced in issue #452 

ajpowelsnl:
This PR adds example code from Compiler Explorer to `MinLoc` built-in reducer.

PR #576, opened 2024-09-03
Request Clang Formatting for Contributions to Main Project & Documentation Examples

ajpowelsnl:
Formatting requirements are not documented.  This PR aims to provide concrete guidance on code formatting vs. general (or no) statements about indentation.  The outstanding question is whether a specific version of `clang-format` would be needed -- a particular version could be a bit of a challenge for external contributors.

### DIFF
--- a/docs/source/templates/class_api.rst
+++ b/docs/source/templates/class_api.rst
@@ -161,7 +161,7 @@ Examples
 
   Prefer working and compilable examples to prose descriptions (such as "Usage").
 
-  Example code should be formatted (in a separate temporary file) using the clang compiler "clang-format -i" utility, and subsequently transferred to documentation entry. 
+  Example code should be formatted (in a separate temporary file) using the clang compiler "clang-format -i" utility, and subsequently transferred to documentation entry.
 
 .. code-block:: cpp
 
--- a/docs/source/templates/class_api.rst
+++ b/docs/source/templates/class_api.rst
@@ -161,6 +161,8 @@ Examples
 
   Prefer working and compilable examples to prose descriptions (such as "Usage").
 
+  Example code should be formatted (in a separate temporary file) using the clang compiler "clang-format -i" utility, and subsequently transferred to documentation entry. 
+
 .. code-block:: cpp
 
   #include <Kokkos_Core.hpp>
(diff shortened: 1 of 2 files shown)


crtrott (changes requested):
The change is actually counter productive: when someone reviews the clang-format stuff is already enforced because it won't pass testing. But there are many style guidelines (such as CamelCase for class names or snake_case for function names) that are not enforced by clang-format, and that is what the reviewer needs to look at.

docs/source/templates/class_api.rst
```diff
@@ -161,6 +161,8 @@ Examples
 
   Prefer working and compilable examples to prose descriptions (such as "Usage").
 
+  Example code should be formatted (in a separate temporary file) using the clang compiler "clang-format -i" utility, and subsequently transferred to documentation entry.
```
crtrott, line 164: Not sure that is necessary, and may be sometimes unhelpful since rendering on the website can often be a constraint by different aspects than rendering for source files.

nmm0: We could just recommend that contributors format code according to Kokkos conventions. (As a side note I think clang-tidy can check naming conventions, but I think this would be too much of a pain for people to run two tools on example code).

ajpowelsnl: The requirement for Kokkos-style indention (currently, from clang-16) was quibbled over in a recent [PR](https://github.com/kokkos/kokkos-core-wiki/pull/556) -- "eye balling" Kokkos-style indentation was not the acceptable response.

On a different, but related point, for example code, we should probably be using the `.. example-code::` [directive](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/code_blocks.html#examples), in addition to the already-instituted `.. code-block:: cpp` directive.

As for the browser determining example rendering, that does not appear to be the case for Firefox and Chrome (formatted example code renders identically and correctly in both).  Kokkos documentation does not render at all in Safari.

**Questions on how / whether to proceed:**

- Benefit in giving more concrete guidance on example formatting requirements, or continue in cycles of quibbling about formatting on per PR basis?
- If being prescriptive about formatting, can `clang-format` be a range of versions (vs. just clang-16)?
- Doesn't warrant further consideration, just close issue?

PR #577, opened 2024-09-03
Add a usecase discussion of how to deal with function pointers

crtrott:


docs/source/usecases/FunctionPointers.md
```diff
@@ -0,0 +1,286 @@
...
+
+This worked on host (both inside and outside a `parallel_for` but crashes inside a device kernel.
+Here is the output:
+
+```
+foo called from host
+I can use the function pointer in a host parallel_for!
+foo called from host
+Now I will crash if we compiled for CUDA/HIP
+cudaDeviceSynchronize() error( cudaErrorIllegalAddress): an illegal memory access was encountered /home/crtrott/Kokkos/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:153
```
JBludau, line 83:
```suggestion
cudaDeviceSynchronize() error( cudaErrorIllegalAddress): an illegal memory access was encountered /kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:153
```

docs/source/usecases/FunctionPointers.md
```diff
@@ -0,0 +1,286 @@
...
+  template<class ... Args>
+  KOKKOS_FUNCTION
+  auto operator()(Args...args) const {
+    KOKKOS_IF_ON_HOST( return h(args...); )
+    KOKKOS_IF_ON_DEVICE( return d(args...); )
+  }
+};
+```
+
+This class is templated on the function pointer type, contains a pointer for both the host and the device version, and has a templated operator that forwards all the arguments, and calls the appropriate function pointer depending on call site.
```
JBludau, line 194:
```suggestion
This class is templated on the function pointer type, contains a pointer for both the host and the device version, and has a templated operator that copies all the arguments, and calls the appropriate function pointer depending on the call site.
```

I find it kind of important to make the distinction here as it adds another drawback.

docs/source/usecases/FunctionPointers.md
```diff
@@ -0,0 +1,286 @@
...
+```c++
+struct SomeClass {
+  DualFunctionPtr<decltype(&foo)> bar;
+  KOKKOS_FUNCTION void print() const {
+    bar();
+  }
+};
+```
+
+However we still need to initialize the two function pointers, using the device initialization approach from before, and also initializing the host side. Note: the order of initialization and `deep_copy` matters, because `deep_copy` will overwrite both member poitners in our wrapper class:
```
JBludau, line 207:
```suggestion
However we still need to initialize the two function pointers using the device initialization approach from before, and also initializing the host side. Note: the order of initialization and `deep_copy` matters, because `deep_copy` will overwrite both member poitners in our wrapper class:
```

docs/source/usecases/FunctionPointers.md
```diff
@@ -0,0 +1,286 @@
...
+    });
+    Kokkos::fence();
+    printf("I didn't crash whoohoo!!\n");
+  }
+  Kokkos::finalize();
+}
+```
+
+While this generally works (and note this code will also work if you simply compile for host-only) there may be architectures in the future where this fails.
+Also we in principle strongly discourage the use of function pointers - the amount of possible problems you run into is fairly significant.
```
JBludau, line 283:
```suggestion
Also we strongly discourage the use of function pointers - the amount of possible problems you run into is fairly significant.
```

I would just make that unconditional

JBludau (commented):
I think it is a good addition

docs/source/usecases/FunctionPointers.md
```diff
@@ -0,0 +1,286 @@
...
+  }
+};
+```
+
+Going forward we will use a simple examplar function:
+
+```c++
+KOKKOS_INLINE_FUNCTION void foo() {
+  KOKKOS_IF_ON_HOST(printf("foo called from host\n");)
+  KOKKOS_IF_ON_DEVICE(printf("foo called from device\n");)
```
masterleinad, line 25:
```suggestion
  KOKKOS_IF_ON_HOST(Kokkos::printf("foo called from host\n");)
  KOKKOS_IF_ON_DEVICE(Kokkos::printf("foo called from device\n");)
```

docs/source/usecases/FunctionPointers.md
```diff
@@ -0,0 +1,286 @@
...
+This function leverages the `KOKKOS_IF_ON_HOST` and `KOKKOS_IF_ON_DEVICE` macros so we can tell which version we got.
+
+Putting it all together into a fully self contained source and lets try to call it both on host and on device:
+
+```c++
+#include <Kokkos_Core.hpp>
+
+KOKKOS_INLINE_FUNCTION void foo() {
+  KOKKOS_IF_ON_HOST(printf("foo called from host\n");)
+  KOKKOS_IF_ON_DEVICE(printf("foo called from device\n");)
```
masterleinad, line 38:
```suggestion
  KOKKOS_IF_ON_HOST(Kokkos::printf("foo called from host\n");)
  KOKKOS_IF_ON_DEVICE(Kokkos::printf("foo called from device\n");)
```

docs/source/usecases/FunctionPointers.md
```diff
@@ -0,0 +1,286 @@
...
+[0x43085d] 
+[0x430a5e] 
+[0x429855] 
+[0x40843e] 
+[0x7f7fca5ba7e5] __libc_start_main
+[0x409cce] 
+Aborted (core dumped)
+```
+
+*The function pointer we created was for a host function: you can not use it on device same as you can't dereference data pointers to host data!*
```
masterleinad, line 96:
```suggestion
*The function pointer we created was for a host function: you can not use it on the device just like you can't dereference data pointers to host data!*
```

docs/source/usecases/FunctionPointers.md
```diff
@@ -0,0 +1,286 @@
...
+    Kokkos::View<SomeClass> A_v("A_v");
+    // Now init the function pointer on device:
+    Kokkos::parallel_for(Kokkos::RangePolicy<Kokkos::DefaultExecutionSpace>(0,1), KOKKOS_LAMBDA(int) {
+      A_v().bar = &foo;
+    });
+    // copy the device instance back to the host
+    Kokkos::deep_copy(A, A_v);
+```
+
+We are leveraging here the fact that `deep_copy` allows you to copy from and to a host scalar value, from a `View` of rank-0.
```
masterleinad, line 116:
```suggestion
We are leveraging here the fact that `deep_copy` allows you to copy in both direction between a  host scalar value and a `View` of rank-0.
```

docs/source/usecases/FunctionPointers.md
```diff
@@ -0,0 +1,286 @@
...
+
+If we do this `A` will contain a function pointer to a device function. That means we can capture `A` into a parallel region and execute it on the device, but now it will crash on the host.
```
masterleinad, line 118:
```suggestion
If we do this, `A` will contain a function pointer to a device function. That means we can capture `A` into a parallel region and execute it on the device, but now it will crash on the host.
```

docs/source/usecases/FunctionPointers.md
```diff
@@ -0,0 +1,286 @@
...
+
+```c++
+#include <Kokkos_Core.hpp>
+#include <cmath>
+
+KOKKOS_INLINE_FUNCTION void foo() {
+  KOKKOS_IF_ON_HOST(printf("foo called from host\n");)
+  KOKKOS_IF_ON_DEVICE(printf("foo called from device\n");)
```
masterleinad, line 126:
```suggestion
  KOKKOS_IF_ON_HOST(Kokkos::printf("foo called from host\n");)
  KOKKOS_IF_ON_DEVICE(Kokkos::printf("foo called from device\n");)
```

docs/source/usecases/FunctionPointers.md
```diff
@@ -0,0 +1,286 @@
...
+```
+Now I can capture A in a device kernel and use it there!
+foo called from device
+But now I will crash on the host :-(
+Segmentation fault (core dumped)
+```
+
+## Creating a dual function pointer object
+
+To do better than this we need something which contains both the device function pointer and the host function pointer.
```
masterleinad, line 176:
```suggestion
To do better than this, we need something which contains both the device function pointer and the host function pointer.
```

docs/source/usecases/FunctionPointers.md
masterleinad, line 207:
```suggestion
However, we still need to initialize the two function pointers, using the device initialization approach from before, and also initializing the host side. Note: the order of initialization and `deep_copy` matters, because `deep_copy` will overwrite both member poitners in our wrapper class:
```

docs/source/usecases/FunctionPointers.md
```diff
@@ -0,0 +1,286 @@
...
+      A.print();
+    });
+    Kokkos::fence();
+    printf("I didn't crash whoohoo!!\n");
+  }
+  Kokkos::finalize();
+}
+```
+
+While this generally works (and note this code will also work if you simply compile for host-only) there may be architectures in the future where this fails.
```
masterleinad, line 282:
```suggestion
While this generally works (and note this code will also work if you simply compile for host-only), there may be architectures in the future where this fails. SYCL (and therefore the SYCL backend) doesn't allow using function pointers on the device for example.
```

PR #578, opened 2024-09-04
Add KOKKOS_FUNCTION etc. documentation

crtrott:


docs/source/API/core/macros-special/host_device_macros.rst
```diff
@@ -0,0 +1,141 @@
...
+        template<class T>
+        KOKKOS_FUNCTION void bar() const { ... }
+    };
+
+    template<class T>
+    KOKKOS_FUNCTION void foo(T v) { ... }
+         
+This macro is also used for non-templated free functions in conjunction with relocatable device code -
+i.e. if one wants to compile functions in some compilation unit A but call them from Kokkos
+parallel constructs defined in compilation unit B.
```
masterleinad, line 50: How is that supposed to work with `SYCL`?

crtrott: "in conjunction" implies its not the only thing you need to do

docs/source/API/core/macros-special/host_device_macros.rst
```diff
@@ -0,0 +1,141 @@
+
+``Function markup macros``
```
dalg24, line 2: Where does the "markup" term come from?
Should we be saying "annotation" instead?

docs/source/API/core/macros-special/host_device_macros.rst
```diff
@@ -0,0 +1,141 @@
...
+i.e. if one wants to compile functions in some compilation unit A but call them from Kokkos
+parallel constructs defined in compilation unit B.
+
+
+``KOKKOS_LAMBDA``
+-----------------
+
+This macro provides default capture clause and host device markup for lambdas. It is the equivalent of
+``[=] __host__ __device__`` in CUDA and HIP.
+It is used than creating C++ lambdas to be passed to Kokkos parallel dispatch mechanisms such as
```
dalg24, line 102: That sentence is weird.
I think that we should consider having some blanket wording that says that any function or lambda called within a kernel is required to be properly annotated.

docs/source/API/core/macros-special/host_device_macros.rst
```diff
@@ -0,0 +1,141 @@
...
+
+.. warning:: Do not use ``KOKKOS_LAMBDA`` inside functions marked as ``KOKKOS_FUNCTION`` etc. or within a lambda marked with ``KOKKOS_LAMBDA``. Specifically do not use ``KOKKOS_LAMBDA`` to define lambdas for nested parallel calls. CUDA does not support that. Use plain C++ syntax instead: ``[=] (int i) {...}``.
+
+.. warning:: When creating lambdas inside of class member functions you may need to use ``KOKKOS_CLASS_LAMBDA`` instead.
+
+``KOKKOS_CLASS_LAMBDA``
+-----------------------
+
+This macro provides default capture clause and host device markup for lambdas created inside of class member functions. It is the equivalent of
+``[=, *this] __host__ __device__`` in CUDA and HIP, capturing the parent class by value instead of by reference.
```
dalg24, line 127: Please add a data member being accessed in the lambda and show that it is an alternative to
```C++
Foo {
...
Data member_;
};

auto const& member = member_;
parallel_for(..., KOKKOS_LAMBDA(...){ ... member ...  });

**Review by dalg24 — COMMENTED** [review]
What about `KOKKOS_DEDUCTION_GUIDE` ?

File: docs/source/API/core/macros-special/host_device_macros.rst (hunk 1 above)
Comment (masterleinad) [reply docs/source/API/core/macros-special/host_device_macros.rst:50]: Hmm... I still find the last sentence more confusing then helpful and would wait with saying anything for functions functions used in separable compilarion/relocatable device code until after discussing https://github.com/kokkos/kokkos/pull/5993.

```suggestion
```

docs/source/API/core/macros-special/host_device_macros.rst
```diff
@@ -0,0 +1,141 @@
...
+        void bar() const { ... }
+    };
+
+    template<class T>
+    KOKKOS_FORCEINLINE_FUNCTION
+    void foo(T v) { ... }
+         
+This macro is also used for non-templated free functions in conjunction with relocatable device code -
+i.e. if one wants to compile functions in some compilation unit A but call them from Kokkos
+parallel constructs defined in compilation unit B.
```
masterleinad, line 94: Similarly here.
```suggestion
```

crtrott: We need some warning here, because otherwise people will pepper everything with the KOKKOS_FUNCTION instead of KOKKOS_INLINE_FUNCTION and it will fail depending on how stuff is set up. This didn't used to be much of a problem because everyone was using KOKKOS_INLINE_FUNCTION by default. But for a while now some team members have strenuously complained about that so folks are moving away from it :-)

docs/source/API/core/macros-special/host_device_macros.rst
crtrott: actually its not because if you generate inside a KOKKOS_FUNCTION you can't use KOKKOS_LAMBDA currently

docs/source/API/core/macros-special/host_device_macros.rst
masterleinad: But currently there is nothing special about relocatable device functions; they behave the same as free functions that might need `KOKKOS_INLINE_FUNCTON`.

crtrott: If you want a function which uses relocatable device code capability you can't mark it KOKKOS_INLINE_FUNCTION. Only if you mark it KOKKOS_FUNCTION will it use that compilation/link technique or?

masterleinad: Yes, it's the same as for a regular `inline` function.

docs/source/API/core/macros-special/host_device_macros.rst
```diff
@@ -0,0 +1,206 @@
...
+            // print_data();
+            // use the copy of data
+            printf("%i %i\n",i,data_copy);
+          });
+        }
+    };
+
+
+``KOKKOS_DEDUCTION_GUIDE``
+-----------------------
```
masterleinad, line 175:
```
/home/runner/work/kokkos-core-wiki/kokkos-core-wiki/docs/source/API/core/macros-special/host_device_macros.rst:175: WARNING: Title underline too short.

``KOKKOS_DEDUCTION_GUIDE``
-----------------------
```
```suggestion
``KOKKOS_DEDUCTION_GUIDE``
--------------------------
```

tpadioleau (approved):

docs/source/API/core/macros-special/host_device_macros.rst
dalg24: Christian pointed out that if any member function is called then we need KOKKOS_CLASS_LAMBDA anyway.
I guess I agree the discussion about data members is not worth it.

PR #584, opened 2024-09-16
Update public headers

tpadioleau:
Following discussion https://github.com/kokkos/kokkos/pull/7320#issuecomment-2353200996, as of Kokkos 4.4, `Kokkos_View.hpp` and `Kokkos_Layout.hpp` are not public headers.


masterleinad (approved):

dalg24 (approved):
Thanks for fixing

PR #590, opened 2024-10-10
team_size_max/recommended: Clarify that team_size should be checked

masterleinad:
Related to https://github.com/kokkos/kokkos/pull/7423.

### DIFF
--- a/docs/source/API/core/policies/TeamPolicy.rst
+++ b/docs/source/API/core/policies/TeamPolicy.rst
@@ -120,7 +120,7 @@ Description
    .. cppkokkos:function:: template<class FunctorType> int team_size_max(const FunctorType& f, const ParallelReduceTag&) const;
 
       Query the maximum team size possible given a specific functor. The tag denotes whether this is for a |parallelFor|_ or a |parallelReduce|_.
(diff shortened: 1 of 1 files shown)


docs/source/API/core/policies/TeamPolicy.rst
```diff
@@ -120,15 +120,15 @@ Description
    .. cppkokkos:function:: template<class FunctorType> int team_size_max(const FunctorType& f, const ParallelReduceTag&) const;
 
       Query the maximum team size possible given a specific functor. The tag denotes whether this is for a |parallelFor|_ or a |parallelReduce|_.
-      Note: this is not a static function! The function will take into account settings for vector length and scratch size of ``*this``. Using a value larger than the return value will result in dispatch failure.
+      Note: this is not a static function! The function will take into account settings for vector length and scratch size of ``*this``. Using a value larger than the return value will result in dispatch failure. If the value returned is non-positive, no valid team size could be found. A common reason is that too much scratch cache memory was requested.
```
dalg24, line 123: Why do you say non-positive rather than "zero"?

masterleinad: We return an `int` and so checking for positive values seems more appropriate. I could see that we would return negative values at some point for particular error cases if we keep a signed return value.

crtrott (approved):

PR #593, opened 2024-10-31
Document deprecation for Kokkos 4

masterleinad:
I gathered everything up to 43a30195af81b6d1aa5b1efec939af8423857f2c. We still need sort them in sections similar to what we are doing for Kokkos 3.

### DIFF
--- a/docs/source/deprecation_page.rst
+++ b/docs/source/deprecation_page.rst
@@ -1,5 +1,54 @@
-Deprecation for Kokkos-3.x
-==========================
+Deprecations
+************
+
+Deprecations for Kokkos 4
+=========================
+
+up to 43a30195af81b6d1aa5b1efec939af8423857f2c
+
+- Default constructor for BinOp1D, BinOp3D
+- Experimental::swap -> kokkos_swap
+- vector
+- ExecutionSpace::in_parallel
+- Cuda::Cuda(cudaStream_t stream) -> Cuda(cudaStream_t stream, Impl::ManageStream manage_stream)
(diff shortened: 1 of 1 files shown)


docs/source/deprecation_page.rst
```diff
@@ -1,5 +1,67 @@
...
+   :header-rows: 1
+
+   * - Deprecated symbol
+     - Since
+     - Replacement
+     - Additional information
+   * - ``Cuda::Cuda(cudaStream_t stream, bool manage_stream)``
+     - 4.2.00
+     - ``Cuda::Cuda(cudaStream_t stream)``
+     - constructing from a Cuda execution space instance should alwyas used an externally managed cudaStream object
```
dalg24, line 20:
```suggestion
     - constructing from a Cuda execution space instance should always used an externally managed ``cudaStream_t`` object
```

dalg24 (commented):
The table does not render super well.
Did you consider alternative approaches, such gathering entries in subsections corresponding to 4.X minor versions and having some bullet list or whatnot with optionally notes for some entries

masterleinad:
It might be enough to just link to the respective pull request for more information.

dalg24:
> It might be enough to just link to the respective pull request for more information.

No it's not.  Even if our PRs systematically came with good descriptions which they don't, that would still put unnecessarily the burden on the user to figure things out.

masterleinad:
I also pushed a list approach. We can discuss tomorrow which one we want to go with.

JBludau (approved):
found some minor things a while ago and forgot to submit

PR #597, opened 2024-11-05
Add documentation for two AMD APUs

Rombur:


docs/source/keywords.rst
```diff
@@ -599,6 +604,11 @@ Kokkos will attempt to autodetect the architecture flag at configuration time.
       * GFX906
       * MI50, MI60
       * (since Kokkos 4.2)
+
+    * * ``Kokkos_ARCH_AMD_GFX1103``
+      * GFX1103
+      * Phoenix APU
```
dalg24, line 610: As I understand it, "Phoenix" is the codename for the microarchitecture whereas this column says "Models".
Should we list a couple popular models, something like "Phoenix APUs (e.g. MODEL-1, MODEL-2, etc.)"

docs/source/keywords.rst
```diff
@@ -575,10 +575,15 @@ Kokkos will attempt to autodetect the architecture flag at configuration time.
...
+    * * ``Kokkos_ARCH_AMD_GFX942_APU``
+      * GFX942
+      * MI300A
+      * (since Kokkos 4.5)
+
     * * ``Kokkos_ARCH_AMD_GFX942``
       * GFX942
       * MI300A, MI300X
-      * (since Kokkos 4.2)
+      * (since Kokkos 4.2, since Kokkos 4.5 this should only be used for MI300X)
```
dalg24, line 586: I haven't looked at how this renders but I expects this folds over multiple lines.
Did it still look fine to you, should we instead do a "see below"?

Rombur: Looks ok to me:
image

dalg24: Sure

docs/source/keywords.rst
Rombur: I have added something

masterleinad (approved):
Fine with me.

PR #601, opened 2024-11-12
Update Compiler Version Requirements for the OpenACC backend

seyonglee:
Update Compiler Version Requirement document (`requirements.rst`) for the OpenACC backend.

### DIFF
--- a/docs/source/requirements.rst
+++ b/docs/source/requirements.rst
@@ -34,6 +34,10 @@ Kokkos 4.x
       * 10.0.0
       * 12.0.0, 14.0.0
 
+    * * Clacc (OpenACC) (experimental)
+      * 40cad5ea5c42
+      * 40cad5ea5c42
+
     * * AppleClang 
       * 8.0
       * latest
@@ -57,6 +61,10 @@ Kokkos 4.x
     * * NVC++ 
       * 22.3
       * 22.9
+
+    * * NVC++ (OpenACC) (experimental) 
+      * 23.7
+      * 24.5
 
     * * ROCM 
       * 5.2.0
--- a/docs/source/requirements.rst
+++ b/docs/source/requirements.rst
@@ -34,10 +34,6 @@ Kokkos 4.x
       * 10.0.0
       * 12.0.0, 14.0.0
 
-    * * Clacc (OpenACC) (experimental)
-      * 40cad5ea5c42
-      * 40cad5ea5c42
-
     * * AppleClang 
       * 8.0
       * latest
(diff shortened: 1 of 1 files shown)


docs/source/requirements.rst
```diff
@@ -34,6 +34,10 @@ Kokkos 4.x
       * 10.0.0
       * 12.0.0, 14.0.0
 
+    * * Clacc (OpenACC) (experimental)
+      * 40cad5ea5c42
+      * 40cad5ea5c42
```
JBludau, line 39: is this a copy/paste error?

seyonglee: No. this is a commit number; Clacc (https://github.com/llvm-doe-org/llvm-project/tree/clacc/main) is a customized version of the upstream LLVM, and thus I used the Clacc's commit number instead of the Clang version number not to be confused with the upstream Clang/LLVM compiler.

masterleinad: I would just drop `Clacc` here. We don't test with it in our `jenkins` CI anyway and if there is no release it's a limited use for the users targeted here.

JBludau: I meant that it lists the same commit hash twice

pzehner: @masterleinad Should it be added to the test matrix, then?

masterleinad: > @masterleinad Should it be added to the test matrix, then?

That's something to discuss.

seyonglee: The entry for Clacc is removed.

docs/source/requirements.rst
```diff
@@ -58,6 +58,10 @@ Kokkos 4.x
       * 22.3
       * 22.9
 
+    * * NVC++ (OpenACC) (experimental) 
+      * 23.7
+      * 24.5
```
dalg24, line 63: I don't think that is accurate.  This column says "primary tested" and we are currently testing with 23.7 in the CI
https://github.com/kokkos/kokkos/blob/cfb48fb7bf1a870a6243125d9cd66cfb0459722c/.jenkins#L185-L196

Are you running nightly builds or such on 24.5?

seyonglee: I locally use 24.5 to test the OpenACC backend using Kokkos unit tests and other Kokkos benchmarks that I have.
As you said, CI uses 23.7, which is the minimum version  that works with the OpenACC backend.
Maybe, I misunderstood what "primary tested" means; I will change the primary tested version to 23.7.

masterleinad (approved):

PR #615, opened 2025-01-22
Adding documentation for Kokkos::Array

nliber:


docs/source/API/containers-index.rst
```diff
@@ -7,6 +7,8 @@ API: Containers
 
    * - Container
      - Description
+   * - `Array <containers/Array.html>`__
```
JBludau, line 10: I see this more like `View` ... do we really want this to be in containers?

docs/source/API/containers/Array.rst
```diff
@@ -0,0 +1,173 @@
...
+
+    = for sections
+
+    - for subsections
+
+    ^ for subsubsections
+
+    " for paragraphs
+
+..
```
JBludau, line 16: did you mean to push the comment stuff?

docs/source/API/containers/Array.rst
```diff
@@ -0,0 +1,173 @@
...
+  .. cppkokkos:function:: constexpr pointer data()
+  .. cppkokkos:function:: constexpr const_pointer data() const
+
+    :return: A pointer to the first element of the array.  If ``N == 0``, the return value is unspecified and not dereferenceable.
+
+
+Deduction Guides
+----------------
+
+.. cppkokkos:function:: template<class T, class... U> Array(T, U...) -> Array<T, 1 + sizeof...(U)>
```
JBludau, line 101: maybe this can be highlighted in an example

docs/source/API/containers/Array.rst
```diff
@@ -0,0 +1,173 @@
...
+
+   :return: ``true`` if and only if ∀ the elements in ``l`` and ``r`` compare equal.
+
+.. cppkokkos:function:: template<class T, size_t N> constexpr bool operator!=(const Array<T, N>& l, const Array<T, N>& r) noexcept
+
+   :return: ``!(l == r)``
+
+.. cppkokkos:function:: template<class T, size_t N> constexpr kokkos_swap(Array<T, N>& l, Array<T, N>& r) noexcept(N == 0 || is_nothrow_swappable_V<T>)
+
+   :return: If ``T`` is swappable or ``N == 0``, each of the elements in `l` and `r` are swapped via ``kokkos_swap``.
```
JBludau, line 120: not really a return ... but might be to render it neatly in sphinx

docs/source/API/containers/Array.rst
```diff
@@ -0,0 +1,173 @@
...
+
+..
+  The (pulic header) file the user will include in their code
+
+Header File: ``Kokkos_Core.hpp``
+
+..
+  High-level, human-language summary of what the thing does, and if possible, brief statement about why it exists (2 - 3 sentences, max);
+
+Description
```
JBludau, line 34: Since we have the old version below: maybe add in which version this one was added?

docs/source/API/containers/Array.rst
```diff
@@ -0,0 +1,173 @@
...
+  .. cppkokkos:type:: reference = T&
+  .. cppkokkos:type:: const_reference = const T&
+  .. cppkokkos:type:: size_type = size_t
+  .. cppkokkos:type:: difference_type = ptrdiff_t
+
+  .. rubric:: Public Member Functions
+
+  .. cppkokkos:function:: static constexpr bool empty()
+
+    :return: ``N == 0``
```
JBludau, line 78: this always returns `false`, right?

docs/source/API/containers-index.rst
nliber: In C++ it is considered a container, but I don't have a strong feeling on it.

docs/source/API/containers/Array.rst
nliber: No; I'll fix it.

docs/source/API/containers/Array.rst
nliber: Passing 0 as the template parameter is perfectly valid, so N == 0 really is the return value (unless I got it wrong, like our implementation of Array had it for years...)

JBludau: https://github.com/kokkos/kokkos/blob/0766bc570e25b7d952d747fffd8c0828b4dcaf34/core/src/Kokkos_Array.hpp#L107
AFAIU we hardcoded `false`

docs/source/API/containers-index.rst
dalg24: Here Containers refers to the Kokkos subpackage, that is, everything that is defined in the `containers/src/` directory.  `Array` belongs to Core so it is not the right place.
The API reference for Core is a bit messy at the moment so I would either put it in Utilities next to `Complex` or in the STL Compat section.  We can put organize things better in follow ups.

docs/source/API/containers/Array.rst
dalg24: There is a specialization for 0-sized arrays which does properly return `true`.
https://github.com/kokkos/kokkos/blob/0766bc570e25b7d952d747fffd8c0828b4dcaf34/core/src/Kokkos_Array.hpp#L175
Apparently, it is not covered in our test though.

docs/source/API/containers-index.rst
nliber: Moved to STL Compat.

docs/source/API/containers/Array.rst
nliber: Yea.  I'm trying to both keep the C++ noise (SFINAE and a separate function definition for `kokkos_swap(Array<T, 0>&, Array<T, 0>&)) out of the function call documentation here, as well as describe the effects of the call.  Unfortunately, I couldn't find a Sphynx keyword other than :return: that lets us render this decently.

docs/source/API/containers/Array.rst
nliber: Done.

docs/source/API/containers/Array.rst
nliber: There is already one in the Examples section.  Should we have more than that?

dalg24 (commented):
Maybe in the preamble say that the intent is to be a drop in replacement for `std::array<T, N>`

dalg24 (approved):

PR #617, opened 2025-01-26
Remove old build notes for cmake <3.12

JBludau:
since we require 3.16 users should not hit this

### DIFF
--- a/docs/source/building.md
+++ b/docs/source/building.md
@@ -25,10 +25,6 @@ When configuring your project just set:
   -DKokkos_ROOT=${kokkos_install_prefix} \
   -DCMAKE_CXX_COMPILER=${compiler_used_to_build_kokkos}
 ````
-Note: You may need the following if using some versions of CMake (e.g. 3.12):
-````cmake
-cmake_policy(SET CMP0074 NEW)
-````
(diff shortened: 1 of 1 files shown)


dalg24 (commented):
I think that was more about what projects downstream set as their cmake minimum required version and that if that is less than 3.12, then `Kokkos_ROOT` might not be the right thing to find Kokkos.

docs/source/building.md
```diff
@@ -25,10 +25,6 @@ When configuring your project just set:
   -DKokkos_ROOT=${kokkos_install_prefix} \
   -DCMAKE_CXX_COMPILER=${compiler_used_to_build_kokkos}
 ````
-Note: You may need the following if using some versions of CMake (e.g. 3.12):
-````cmake
-cmake_policy(SET CMP0074 NEW)
-````
```
masterleinad, line 31: Note: Using `Kokkos_ROOT` requires [CMP0074](https://cmake.org/cmake/help/latest/policy/CMP0074.html) to be set to `NEW`, e.g., using
````cmake
cmake_policy(SET CMP0074 NEW)
````
when the minimum required CMake version in the dowsntream project is less than 3.12.

JBludau:
Not sure if we really test this, but I will close this for now. If we update our cmake build system we will need to revisit this because there we impose 3.21 on downstream projects.

PR #624, opened 2025-02-01
Improve the Get Started section

dalg24:
Users should be able to get their first program up and running after reading this section.

### DIFF
--- a/docs/source/quick_start.rst
+++ b/docs/source/quick_start.rst
@@ -4,83 +4,171 @@ Quick Start
 This guide is intended to jump start new Kokkos users (and beginners, in particular).
 
 
-Download Latest and Build 
------------------------------
+Prerequisites
+~~~~~~~~~~~~~
+
+To complete this tutorial, you'll need:
+
+* a compatible operating system (e.g. Linux, macOS, Windows).
+* a compatible C++ compiler that supports at least C++17.
+* `CMake <https://cmake.org/>`_ and a compatible build tool for building the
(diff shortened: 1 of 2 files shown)


docs/source/quick_start.rst
```diff
@@ -4,83 +4,171 @@ Quick Start
...
+project. You’ll use this file to set up your project and declare a dependency
+on Kokkos.
+
+First, create a directory for your project:
 
 .. code-block:: sh
+
+  > mkdir MyProject && cd MyProject
+
+Next, you’ll create the ``CMakeLists.txt`` file and declare a dependency on
```
JBludau, line 44: maybe redundant with the line 48

docs/source/quick_start.rst
```diff
@@ -4,83 +4,171 @@ Quick Start
...
------------------------------
+Prerequisites
+~~~~~~~~~~~~~
 
-.. note::
+To complete this tutorial, you'll need:
+
+* a compatible operating system (e.g. Linux, macOS, Windows).
+* a compatible C++ compiler that supports at least C++17.
+* `CMake <https://cmake.org/>`_ and a compatible build tool for building the
```
JBludau, line 14: maybe spell out the minimum version here? we probably require one implicitly ... but might be outside of the scope here

JBludau (approved):
the suggestions are def. non-blocking

crtrott (approved):

PR #623, opened 2025-01-30
Update action to deploy to GitHub pages

dalg24:
This is an attempt to fix the permissions error on the last merge. The code is inspired from https://github.com/actions/upload-pages-artifact

PR #625, opened 2025-02-03
Allow writing contents when deploying

masterleinad:
Alternative to https://github.com/kokkos/kokkos-core-wiki/pull/623. Fixes deplotibng, also see https://github.com/JamesIves/github-pages-deploy-action?tab=readme-ov-file#getting-started-airplane. The `GITHUB_TOKEN` used only has read permissions and we have to explicitly grant write permissions in this workflow.

### DIFF
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -26,6 +26,7 @@ jobs:
         working-directory: ${{ env.docs-directory }}
         run: |
           make html
+      # .nojekyll file is needed for GitHub Pages to know it's getting a ready webpage
       # and there is no need to generate anything
       - name: Generate nojekyll file
         working-directory: ${{ env.docs-directory }}/generated_docs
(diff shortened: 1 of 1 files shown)


JBludau:
could you explain the differences between the two approaches?

masterleinad:
> could you explain the differences between the two approaches?

This one is very explicit about what it's doing, i.e., copy files from one dircetory to a specific branch. The other one (#623) deploys to Github pages and relies more on the settings in the repository to do what is necessary so it's more difficult to debug. The error I was running into last can be found in https://github.com/masterleinad/kokkos-core-wiki/actions/runs/13119609186.

.github/workflows/deploy_docs.yml
```diff
@@ -26,21 +26,34 @@ jobs:
         working-directory: ${{ env.docs-directory }}
         run: |
           make html
-      - name: Archive documentation
-        uses: actions/upload-artifact@v4
-        with:
-          name: documentation.tar.gz
-          path: ${{ env.docs-directory }}/generated_docs
-      # .nojekyll file is needed for GitHub Pages to know it's getting a ready webpage
```
JBludau, line 34: hmm did you accidentally delete half the comment?

masterleinad: Fixed.

PR #631, opened 2025-02-17
Major overhaul of build instructions and reorganizing sections

dalg24:


docs/source/get-started/building-from-source.rst
```diff
@@ -0,0 +1,235 @@
...
+different name if you prefer).  Kokkos requires out-of-source builds.  The
+``[<options...>]`` part is where you specify the configuration options.
+
+**Common CMake Options**
+
+These options are generally useful for any CMake project:
+
+* ``-DCMAKE_CXX_COMPILER=<compiler>``: Specifies the full path to the C++
+  compiler. For example, use ``hipcc`` for AMD GPUs, ``icpx`` for Intel GPUs,
+  or ``g++`` or ``clang++`` for CPUs.
```
crtrott, line 126: We may wanna say something about CUDA here. There are so many examples, but NVIDIA is missing. A user could assume that simply means he should set NVCC as the CMAKE_CXX_COMPILER since one uses the vendor compilers for the AMD and Intel GPUs.

docs/source/tutorials-and-examples.rst
```diff
@@ -0,0 +1,35 @@
...
+Kokkos organization on GitHub.  Here is the `link
+<https://github.com/kokkos/kokkos-tutorials>`_.
+
+:doc:`Video Lectures <tutorials-and-examples/video-lectures>`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+A series of video lectures on Kokkos was recorded in 2020 as part of an
+Exascale Computing Project (ECP) initiative during the COVID-19 pandemic.
+These lectures provide a comprehensive introduction to Kokkos and cover many of
+its key features. While some details might be slightly outdated due to ongoing
+development, the core concepts and techniques remain relevant.  The lectures
```
crtrott, line 25:
```suggestion
development, most of the content is fully applicable even today.  The lectures
```

crtrott (approved):
Two suggestions/comments. But I am ok with this merging as is and then fixing. Or you can fix it and then just merge.

docs/source/get-started/building-from-source.rst
dalg24: Let's improve in a follow up

PR #635, opened 2025-02-26
Spack section

cedricchevalier19:
Try explain:
- how to compile with Spack
- how to make a Kokkos dependent package

### DIFF
--- a/docs/source/get-started/package-managers.rst
+++ b/docs/source/get-started/package-managers.rst
@@ -47,15 +47,15 @@ To install Kokkos with Spack with default options, run:
     spack install kokkos
 
 
-To install Kokkos with Cuda backend enabled, run:
+To install Kokkos with CUDA backend enabled, run:
 
 .. code-block::
 
-    spack install kokkos +cuda cuda_arch=86
+    spack install kokkos +cuda cuda_arch=90
 
 
-Note that the `cuda_arch` option is specific to the target GPU architecture.  Here, the `cuda_arch` value `86` corresponds
-to the NVIDIA Ampere architecture. With Spack, the architecture must be specified explicitly (no auto-detection).
+Note that the `cuda_arch` option is specific to the target GPU architecture.  Here, the `cuda_arch` value `90` corresponds
+to the NVIDIA Hopper architecture. With Spack, the architecture must be specified explicitly (no auto-detection).
 
 
 For AMD GPU, the traditional Spack's keyword is `rocm` instead of `hip` in Kokkos' CMake. So to install Kokkos with the HIP backend enable, run:
@@ -98,8 +98,7 @@ You have to create a recipe that is contained in a file named `package.py` in th
 You can package as usual with Spack (for example, you can follow the `packaging tutorial <https://spack-tutorial.readthedocs.io/en/latest/tutorial_packaging.html>`_),
 but you have to take into account that for certain backends, you might need to specify the compiler to use.
 
-The compiler might be a wrapper, like `nvcc_wrapper` for CUDA. It is exported by the Kokkos package as the `kokkos_cxx` attribute
-(Currently only available on the develop branch of Spack).
+The compiler might be a wrapper, like `nvcc_wrapper` for CUDA. It is exported by the Kokkos package as the `kokkos_cxx` attribute.
 
 Here is an example of a `package.py` file that includes Kokkos as a dependency:
 


docs/source/get-started/package-managers.rst
```diff
@@ -6,19 +6,100 @@ Use your favorite package manager to install Kokkos.
...
+            args.append(self.define("CMAKE_CXX_COMPILER", self["kokkos"].kokkos_cxx))
+            return args
+
+
+For more complete examples, you can look at already existing recipes in the *Required by* section of
+`Kokkos Spack recipe <https://packages.spack.io/package.html?name=kokkos>`_ or by running:
+
+.. code-block::
+
+    spack dependents kokkos
```
tpadioleau, line 104: Nice I did not know that one!

docs/source/get-started/package-managers.rst
```diff
@@ -6,19 +6,100 @@ Use your favorite package manager to install Kokkos.
...
+
+Most of the time, Spack Kokkos' variants follow the same options as the Kokkos `CMake options <./configuration-guide.html>`_.
+List of available variants can be found by running
+
+.. code-block::
+
+    spack info kokkos
+
+
+Note that CPU micro-architectures are automatically inherited from Spack's configuration.
```
dalg24, line 34: I don't quite understand that sentence.
Does that mean Spack is auto-detecting your CPU arch and passing flags to Kokkos?

docs/source/get-started/package-managers.rst
```diff
@@ -6,19 +6,100 @@ Use your favorite package manager to install Kokkos.
...
+
+To install Kokkos with Cuda backend enabled, run:
+
+.. code-block::
+
+    spack install kokkos +cuda cuda_arch=86
+
+
+Note that the `cuda_arch` option is specific to the target GPU architecture.  Here, the `cuda_arch` value `86` corresponds
+to the NVIDIA Ampere architecture. With Spack, the architecture must be specified explicitly (no auto-detection).
```
dalg24, line 54: Is there similar advice we should give for HIP?
Users might be confused by the fact Spack named the variant `rocm`.
What about SYCL?

cedricchevalier19: It is true that `rocm` instead of `hip` is not straightforward. I will add examples for AMD and Intel.

tpadioleau (commented):
That looks already good to me. I have a question regarding how far should we go:
- do we explain other packages like Kokkos Kernels, Kokkos-fft, Kokkos Tools... ? Maybe only Kokkos Tools has some specificity ?
- I see you mention the develop branch. Maybe we should write the text having spack version 1 in mind ?

docs/source/get-started/package-managers.rst
tpadioleau: Yes my understanding spack auto detects it to give a default value to `arch`. Then the Kokkos recipe does the mapping to the corresponding Kokkos `ARCH` option.

As @cedricchevalier19  said, if needed you can manually specify `arch`.

docs/source/get-started/package-managers.rst
```diff
@@ -6,19 +6,100 @@ Use your favorite package manager to install Kokkos.
...
+        version('1.0', git='foo.git', tag='v1.0')
+
+        depends_on('kokkos')
+
+        def cmake_args(self):
+            args = []
+            # Ensure that the proper compiler is used
+            # It might be nvcc_wrapper
+            args.append(self.define("CMAKE_CXX_COMPILER", self["kokkos"].kokkos_cxx))
+            return args
```
tpadioleau, line 96:
```suggestion
    from spack.package import *

    class Foo(CMakePackage):
        # Usual description of a Spack package

        depends_on("kokkos")

        def cmake_args(self):
            args = []
            # Ensure that the proper compiler is used
            # It might be nvcc_wrapper
            args.append(self.define("CMAKE_CXX_COMPILER", self["kokkos"].kokkos_cxx))
            return args
```

What do you think of just defining what is necessary to illustrate the compiler trick ?

cedricchevalier19: Great idea

Rombur (approved):

docs/source/get-started/package-managers.rst
```diff
@@ -6,19 +6,124 @@ Use your favorite package manager to install Kokkos.
...
+
+To install Kokkos with Cuda backend enabled, run:
+
+.. code-block::
+
+    spack install kokkos +cuda cuda_arch=86
+
+
+Note that the `cuda_arch` option is specific to the target GPU architecture.  Here, the `cuda_arch` value `86` corresponds
+to the NVIDIA Ampere architecture. With Spack, the architecture must be specified explicitly (no auto-detection).
```
tpadioleau, line 58:
```suggestion
    spack install kokkos +cuda cuda_arch=90


Note that the `cuda_arch` option is specific to the target GPU architecture.  Here, the `cuda_arch` value `90` corresponds
to the NVIDIA Hopper architecture. With Spack, the architecture must be specified explicitly (no auto-detection).
```
Should we use the latest flags ?

docs/source/get-started/package-managers.rst
```diff
@@ -6,19 +6,124 @@ Use your favorite package manager to install Kokkos.
...
+++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+Let's say you have a project `Foo` that depends on `Kokkos`. You can package your project with Spack and include `Kokkos` as a dependency.
+
+You have to create a recipe that is contained in a file named `package.py` in the `foo` directory of your recipe repository.
+You can package as usual with Spack (for example, you can follow the `packaging tutorial <https://spack-tutorial.readthedocs.io/en/latest/tutorial_packaging.html>`_),
+but you have to take into account that for certain backends, you might need to specify the compiler to use.
+
+The compiler might be a wrapper, like `nvcc_wrapper` for CUDA. It is exported by the Kokkos package as the `kokkos_cxx` attribute
+(Currently only available on the develop branch of Spack).
```
tpadioleau, line 102:
```suggestion
The compiler might be a wrapper, like `nvcc_wrapper` for CUDA. It is exported by the Kokkos package as the `kokkos_cxx` attribute.
```

I am not confident of how this sentence will evolve :)

tpadioleau (approved):

dalg24:
Thank you!

PR #637, opened 2025-03-11
Document GCC Quad-Precision Math Library support

dalg24:
cc @csiefer2

### DIFF
--- a/docs/source/get-started/configuration-guide.rst
+++ b/docs/source/get-started/configuration-guide.rst
@@ -300,6 +300,10 @@ The following options control enabling TPLs:
       * Whether to enable the LIBDL library
       * ``ON``
       *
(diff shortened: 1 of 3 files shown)


docs/source/get-started/configuration-guide.rst
```diff
@@ -300,6 +300,10 @@ The following options control enabling TPLs:
       * Whether to enable the LIBDL library
       * ``ON``
       *
+    * * ``Kokkos_ENABLE_LIBQUADMATH``
+      * Whether to enable support for 128-bit floating-point types via the GCC Quad-Precision Math Library
```
dalg24, line 304:
```suggestion
      * Whether to enable support for 128-bit floating-point types via GCC's Quad-Precision Math Library 
```

tcclevenger (approved):

PR #646, opened 2025-03-18
Add initial MultiGPU documentation

tcclevenger:
Multi GPU was not yet documented, seems like we needed something for this.

There is much more that could be added
- timings
- show how to alloc unmanaged views
- show how to use the execution space for view alloc, policy creation, fencing
- link the tutorial?

Let me know your thoughts.

### DIFF
--- a/docs/source/API/core/MultiGPUSupport.rst
+++ b/docs/source/API/core/MultiGPUSupport.rst
@@ -151,4 +151,4 @@ Example:
 Notes
 -----
 
-- A `tutorial <https://github.com/kokkos/kokkos-tutorials/tree/main/Exercises/multi_gpu_cuda>`_ for using multi-gpu on CUDA is available.
+- A `tutorial <https://github.com/kokkos/kokkos-tutorials/tree/main/Exercises/multi_gpu_cuda>`_ for using multi-GPU on CUDA is available.
--- a/docs/source/API/core/MultiGPUSupport.rst
+++ b/docs/source/API/core/MultiGPUSupport.rst
@@ -18,7 +18,9 @@ CUDA
 
 For CUDA backend, the user creates a ``cudaStream_t`` object and passes it to the ``Kokkos::Cuda`` constructor.
 
-.. note:: The user is expected to manage the lifetime of the stream. The related execution space needs to be destroyed before destroying the stream.
+.. note:: The lifetime of all Kokkos objects related to the stream (e.g., the execution space) needs to end before the stream itself is invalidated
+          via ``cudaStreamDestroy``. This can be done by adding a scope around such objects, as is done below.
+
 
 .. code-block:: cpp
 
@@ -53,7 +55,8 @@ HIP
 
 For the HIP backend, like with CUDA, the user creates a ``hipStream_t`` object and passes it to the ``Kokkos::HIP`` constructor.
 
-.. note:: The user is expected to manage the lifetime of the stream. The related execution space needs to be destroyed before destroying the stream.
+.. note:: The lifetime of all Kokkos objects related to the stream (e.g., the execution space) needs to end before the stream itself is invalidated
+          via ``cudaStreamDestroy``. This can be done by adding a scope around such objects, as is done below.
 
 .. warning:: Multi-GPU only supported for ROCm 5.6 and later. Because of the lack of HIP API functions for querying a
              stream's device before ROCm 5.6, constructing a ``Kokkos::HIP`` instance on a non-default device isn't
--- a/docs/source/API/core/MultiGPUSupport.rst
+++ b/docs/source/API/core/MultiGPUSupport.rst
@@ -127,8 +127,7 @@ Example:
 
 .. code-block:: cpp
 
-    using ExecutionSpace = decltype(exec_space);
-    Kokkos::View<int*, ExecutionSpace> V(Kokkos::view_alloc("V", exec_space), 10);
+    Kokkos::View<int*> V(Kokkos::view_alloc("V", exec_space), 10);
 
 Launching Kernels
 ~~~~~~~~~~~~~~~~~
--- a/docs/source/API/core/MultiGPUSupport.rst
+++ b/docs/source/API/core/MultiGPUSupport.rst
@@ -1,0 +1,117 @@
+.. role:: cppkokkos(code)
+    :language: cppkokkos
+
+Multi-GPU Support
+=================
+
+Kokkos has support for launching kernels on multiple GPU devices within a single node from host on a single MPI rank for the Cuda, HIP, and SYCL backends.
+
+Using this feature requires knowledge of backend specific API calls for setting devices, creating streams, and (potentially)
+allocating device specific data.
+
+Basic Usage
+-----------
+
+For allocating and launching kernels on a chosen device, the user is responsible for creating device specific streams to pass to backend constructor.
+Once these are created, they can be passed as any other execution space instance to create policies, allocated managed views, asynchronous deep copies, etc. For creating unmanaged
+views, the user must manually allocate device specific memory to pass to the view constructor.
+
+Contructing Execution Space
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Cuda
+^^^^
+
+.. code-block:: cpp
+
+    // Query number of devices available
+    int n_devices;
+    cudaGetDeviceCount(&n_devices);
+
+    // Choose some 0 <= N < n_devices
+    int N = ...;
+
+    // Create stream on device N
+    cudaStream_t stream;
+    cudaSetDevice(N);
+    cudaStreamCreate(&stream);
+
+    // Scope execution space to ensure stream
+    // is destroyed *after* execution space
+    {
+      // Create execution space
+      Kokkos::Cuda exec_space(stream);
+
+      // Use execution space
+      /* ... */
+    }
+
+    // Destroying stream after use
+    cudaSetDevice(N);
+    cudaStreamDestroy(stream);
+
+HIP
+^^^
+
+.. warning:: Multi-GPU only supported for rocm 5.6 and later. Because of the lack of HIP API functions for querying a stream's device
+             in pre rocm 5.6, constructing a ``Kokkos::HIP`` instance on the non-default device will not be able to throw an error,
+             problems will just likely arise downstream.
+
+
+.. code-block:: cpp
+
+    // Query number of devices available
+    int n_devices;
+    hipGetDeviceCount(&n_devices);
+
+    // Choose some 0 <= N < n_devices
+    int N = ...;
+
+    // Create stream on device N
+    hipStream_t stream;
+    hipSetDevice(N);
+    hipStreamCreate(&stream);
+
+    // Scope execution space to ensure stream
+    // is destroyed *after* execution space
+    {
+      // Create execution space
+      Kokkos::HIP exec_space(stream);
+
+      // Use execution space
+      /* ... */
+    }
+
+    // Destroying stream after use
+    hipSetDevice(N);
+    hipStreamDestroy(stream);
+
+SYCL
+^^^^
+
+.. code-block:: cpp
+
+    // Get list of devices available
+    std::vector<sycl::device> gpu_devices =
+      sycl::device::get_devices(sycl::info::device_type::gpu);
+
+    // Choose some 0 <= N < gpu_devices.size()
+    int N = ...;
+
+    // Create a queue on device N.
+    // Note: Kokkos requires SYCL queues to be "in_order"
+    sycl::queue queue{gpu_devices[N], sycl::property::queue::in_order()};
+
+    // Create execution space
+    Kokkos::SYCL exec_space(queue);
+
+    // Use execution space
+    /* ... */
+
+Notes
+-----
+
+- For ``parallel_reduce(..., const ReducerArgument&... reducer)``, it is important to pass rank-0 views for the ``reducer`` arguments
+  instead of scalar types. Scalar reductions imply a fence on all execution spaces.
+- When passing a stream to an execution space constructor, the user is expected the manage the lifetime of the stream. Only after the
+  execution space is destroyed can the user safely destroy the stream.
--- a/docs/source/API/core/MultiGPUSupport.rst
+++ b/docs/source/API/core/MultiGPUSupport.rst
@@ -4,7 +4,8 @@
 Multi-GPU Support
 =================
 
-Kokkos has support for launching kernels on multiple GPU devices within a single node from host on a single MPI rank for the CUDA, HIP, and SYCL backends.
+Kokkos has support for launching kernels on multiple GPUs from a single host process, e.g., a single MPI rank. This feature currently
+exists for the CUDA, HIP, and SYCL backends.
 
 Using this feature requires knowledge of backend specific API calls for creating non-default execution space instances. Once the execution space has been
 created, it can be used to create execution policies, allocate views, etc. all on the device chosen by the user.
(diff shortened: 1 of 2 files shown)


docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,117 @@
...
+    Kokkos::SYCL exec_space(queue);
+
+    // Use execution space
+    /* ... */
+
+Notes
+-----
+
+- For ``parallel_reduce(..., const ReducerArgument&... reducer)``, it is important to pass rank-0 views for the ``reducer`` arguments
+  instead of scalar types. Scalar reductions imply a fence on all execution spaces.
```
masterleinad, line 115: I don't think we promise a global fence but only that the results is available after the call. In any case, this doesn't seem to be specific for multi-GPU setups so I would just remove.

masterleinad:
> There is much more that could be added
> * timings
> * show how to alloc unmanaged views
> * show how to use the execution space for view alloc, policy creation, fencing
> * link the tutorial?

I think we should add showing how to allocate data on the desired device using managed Views. I don't think there is anything specific to unamanaged Views. Linking the tutorial is also a good idea.

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
+.. role:: cppkokkos(code)
+    :language: cppkokkos
+
+Multi-GPU Support
+=================
+
+Kokkos has support for launching kernels on multiple GPU devices within a single node from host on a single MPI rank for the CUDA, HIP, and SYCL backends.
```
JBludau, line 7:
```suggestion
Kokkos has support for launching kernels on multiple GPU devices from a single host process e.g. a single MPI rank. This feature currently exists for the CUDA, HIP, and SYCL backends.
```

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
...
+Constructing Execution Spaces
+-----------------------------
+
+CUDA
+~~~~
+
+For CUDA backend, the user creates a ``cudaStream_t`` object and passes it to the ``Kokkos::Cuda`` constructor.
+
+.. note:: The user is expected to manage the lifetime of the stream. Only after the execution space is destroyed
+          can the user safely destroy the stream.
```
JBludau, line 21:
```suggestion
.. note:: The user is expected to manage the lifetime of the stream. The related execution space needs to be destroyed before destroying the stream.
```

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
...
+
+    // Scope execution space to ensure stream
+    // is destroyed *after* execution space
+    {
+      // Create execution space
+      Kokkos::Cuda exec_space(stream);
+
+      // Use execution space
+      /* ... */
+    }
```
JBludau, line 46: When do we initialize? Since objects are destroyed in the reverse order of creation, we would not need a scope to enforce that, right?

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
...
+    cudaSetDevice(N);
+    cudaStreamDestroy(stream);
+
+HIP
+~~~
+
+For the HIP backend, like with CUDA, the user creates a ``hipStream_t`` object and passes it to the ``Kokkos::HIP`` constructor.
+
+.. note:: The user is expected to manage the lifetime of the stream. Only after the execution space is destroyed
+          can the user safely destroy the stream.
```
JBludau, line 58:
```suggestion
.. note:: The user is expected to manage the lifetime of the stream. The related execution space needs to be destroyed before destroying the stream.
```

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
...
+
+.. warning:: Multi-GPU only supported for ROCm 5.6 and later. Because of the lack of HIP API functions for querying a
+             stream's device before ROCm 5.6, constructing a ``Kokkos::HIP`` instance on a non-default device isn't
+             supported, and problems will likely arise downstream.
```
JBludau, line 62:
```suggestion
             supported.
```
I would not want to get into the details

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
...
+    // Choose some 0 <= N < n_devices
+    int N = ...;
+
+    // Create stream on device N
+    hipStream_t stream;
+    hipSetDevice(N);
+    hipStreamCreate(&stream);
+
+    // Scope execution space to ensure stream
+    // is destroyed *after* execution space
```
JBludau, line 80: same question as above

docs/source/API/core/MultiGPUSupport.rst
JBludau: maybe we should just move the note into the general part since we repeat it otherwise, right?

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
...
+
+    // Destroying stream after use
+    hipSetDevice(N);
+    hipStreamDestroy(stream);
+
+SYCL
+~~~~
+
+For the SYCL backend, the user creates a ``sycl::queue`` object and passes it to the ``Kokkos::SYCL`` constructor.
+
```
JBludau, line 97: We should also add the note here ... or move it to a different place for all of them.

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
...
+    // Use execution space
+    /* ... */
+
+Using Kokkos Methods
+--------------------
+
+Once an execution space has been created on the chosen device, the execution space must be passed to all Kokkos methods
+intended to be used on device. In the absence of an execution space, Kokkos will use the execution space on the default
+device.
```
JBludau, line 122:
```suggestion
intended to be used on the chosen device. If no execution space is passed, Kokkos will use `DefaultExecutionSpace`.
```

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
...
+Launching Kernels
+~~~~~~~~~~~~~~~~~
+
+To launch a kernel on device, pass the execution space to the policy constructor.
+
+Example:
+
+.. code-block:: cpp
+
+    Kokkos::parallel_for("inc_V", Kokkos::RangePolicy<ExecutionSpace>(exec_space, 0, 10),
```
JBludau, line 145:
```suggestion
    Kokkos::parallel_for("inc_V", Kokkos::RangePolicy(exec_space, 0, 10),
```
do we not have CTAD for this?

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+To allocate a managed view on device, pass the execution space to ``Kokkos::view_alloc()``.
+
+Example:
+
+.. code-block:: cpp
+
+    using ExecutionSpace = decltype(exec_space);
+    Kokkos::View<int*, ExecutionSpace> V(Kokkos::view_alloc("V", exec_space), 10);
```
JBludau, line 134:
```suggestion
    auto V(Kokkos::view_alloc("V", exec_space), 10);
```

JBludau (commented):
minor things

docs/source/API/core/MultiGPUSupport.rst
tcclevenger: We have to destroy the stream below before the execution space would go out of scope, assuming they are in the same function.

Do you mean initialize kokkos? We could do that any time before creating the execution space.

This is how the tutorial looks 
```cpp

cudaStreamCreate(&stream);

Kokkos::initialize(...);
{
  Kokkos::Cuda exec_space(stream);

  ...
}
Kokkos::finalize();

cudaStreamDestroy(stream);
```
I could be misunderstanding the question.

docs/source/API/core/MultiGPUSupport.rst
tcclevenger: I thought `sycl::queue` did not need to be managed manually? I'm a bit out of my depth here though, maybe @masterleinad can back me up or refute..

masterleinad: Yes, `sycl::queue` behaves like a `std::shared_ptr`.

docs/source/API/core/MultiGPUSupport.rst
JBludau: yeah, you are right. But this should work:
https://godbolt.org/z/MhY5n3Pbq

docs/source/API/core/MultiGPUSupport.rst
tcclevenger: It's not because of `stream` being deleted after the execution space, it's that we're required to manually destroy the stream before it's deleted.

But maybe then it wouldn't matter if we call `cudaStreamDestroy(stream)` even while the execution space existed? It would be awkward though if we had an execution space instance floating around with an invalid stream.

masterleinad (dismissed):
Looks good enough for a first version. We can improve based on user feedback.

docs/source/API/core/MultiGPUSupport.rst
masterleinad: I'd rather be explicit about the template arguments (and I like passing memory spaces as template arguments to `Kokkos::View` instead of execution spaces but I don't think that's crucial here).

docs/source/API/core/MultiGPUSupport.rst
JBludau: oh, ok. "Lifetime" is the word that triggered me then. For me "Lifetime" in c++ is the time from creation until destruction (destructor call) of the class instance. IIUC you mean the lifetime of the execution space needs to end before we destroy (aka invalidate) the stream before the lifetime of the stream can end. Is that right?

tcclevenger: Yeah, that's right!

JBludau: ok, in that case we definitely need the scope. Maybe we can make it more expressive, e.g.:

.. note:: The user needs to manually manage the lifetime of Kokkos objects that use the stream. The lifetime of all Kokkos objects including the execution space that are related to a stream needs to end before the stream is invalidated via `cudaStreamDestroy`. This can be done via adding a scope as done in the example below.

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,149 @@
...
+Allocating Managed Views
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+To allocate a managed view on device, pass the execution space to ``Kokkos::view_alloc()``.
+
+Example:
+
+.. code-block:: cpp
+
+    Kokkos::View<int*> V(Kokkos::view_alloc("V", exec_space), 10);
```
masterleinad, line 130: I don't agree with removing the `ExecutionSpace` here.
```suggestion
    ExecutionSpace exec_space(...);
    Kokkos::View<int*, typename ExecutionSpace::memory_space> V(Kokkos::view_alloc("V", exec_space), 10);
```

JBludau: can you explain why?

masterleinad: It matters what the type of the execution space is. It has to match with the memory space. I'm fine with using the defualt one here for it to read.
```suggestion
    Kokkos::DefaultExecutionSpace exec_space(...);
    Kokkos::View<int*> V(Kokkos::view_alloc("V", exec_space), 10);
```
but still prefer using a generic one.

JBludau: I agree, that spelling the executionspace out in the example is a good idea. But don't we want to imply that these Views should be created with the exec space that is associated to a stream? This exec space was called `exec_space` in the snippets above.
Does this make sense?

masterleinad: I'm not arguing about the spelling of `exec_space` and we can be more explicit by adding a comment
```suggestion
    ExecutionSpace exec_space(...); // constructed as above
    Kokkos::View<int*, typename ExecutionSpace::memory_space> V(Kokkos::view_alloc("V", exec_space), 10);
```

docs/source/API/core/MultiGPUSupport.rst
tcclevenger: Updated the note, let me know what you think.

docs/source/API/core/MultiGPUSupport.rst
tcclevenger: Updated by querying the type from execution space. I agree with Jakob, I was assuming the reader would follow `exec_space` through the doc.

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
...
+    cudaSetDevice(N);
+    cudaStreamDestroy(stream);
+
+HIP
+~~~
+
+For the HIP backend, like with CUDA, the user creates a ``hipStream_t`` object and passes it to the ``Kokkos::HIP`` constructor.
+
+.. note:: The lifetime of all Kokkos objects related to the stream (e.g., the execution space) needs to end before the stream itself is invalidated
+          via ``cudaStreamDestroy``. This can be done by adding a scope around such objects, as is done below.
```
JBludau, line 59:
```suggestion
.. note:: The lifetime of all Kokkos objects related to the stream (e.g., the execution space) needs to end before the stream itself is invalidated
          via ``hipStreamDestroy``. This can be done by adding a scope around such objects, as is done below.
```
I guess a copy-pasta error

docs/source/API/core/MultiGPUSupport.rst
JBludau: I like it

JBludau (approved):
Looks good to me

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
...
+    Kokkos::SYCL exec_space(queue);
+
+    // Use execution space
+    /* ... */
+
+Using Kokkos Methods
+--------------------
+
+Once an execution space has been created on the chosen device, the execution space must be passed to all Kokkos methods
+intended to be used on the chosen device. If no execution space is passed, Kokkos will use `DefaultExecutionSpace`.
```
masterleinad, line 122:
```suggestion
intended to be used on the chosen device. If no execution space is passed, Kokkos will use the default execution space instance associated with the device Kokkos was initialized with.
```

docs/source/API/core/MultiGPUSupport.rst
```diff
@@ -0,0 +1,153 @@
...
+
+    Kokkos::parallel_for("inc_V", Kokkos::RangePolicy(exec_space, 0, 10),
+      KOKKOS_LAMBDA (const int i) {
+        V(i) += i;
+    });
+
+Notes
+-----
+
+- A `tutorial <https://github.com/kokkos/kokkos-tutorials/tree/main/Exercises/multi_gpu_cuda>`_ for using multi-gpu on CUDA is available.
```
masterleinad, line 153:
```suggestion
- A `tutorial <https://github.com/kokkos/kokkos-tutorials/tree/main/Exercises/multi_gpu_cuda>`_ for using multi-GPU on CUDA is available.
```

masterleinad (approved):

PR #653, opened 2025-04-08
Fix two broken links in the cmake keywords page

tretre91:
This PR fixes two broken links in the CMake keywords page


nmm0 (approved):
Looks good!

PR #655, opened 2025-04-09
doc: remove section numbering

romintomasetti:
This PR is a proposal to remove the manual section numbering from the programming guide.

Reasons to do this:
1. It's easy to number a chapter, but it's really a pain to number the sub(sub(sub))sections.
2. I spotted numbering issues (wrong numbering) and inconsistencies (sometimes subsubsection were not numbered).
3. A reordering of the chapters would mean a manual renumbering procedure.
4. The programming guide was the only "big section" to get section numbering. It's not the case for any other like API or quick start).

Reasons not to do this:
1. The numbering guides the beginner through the reading?

If the consensus is that we want the numbering, I suggest we go the [automated way with a `Sphinx` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-option-toctree-numbered):
```rst
.. toctree::
   :numbered:

   foo
   bar
```


docs/source/ProgrammingGuide/View.rst
```diff
@@ -94,17 +94,17 @@ The requirement that the destructor of `T` not deallocate memory technically dis
...
 .. warning::
 
   NOVICES: THE ANSWER FOR YOU IS "NO."  PLEASE SKIP THIS SECTION.  
 
 A "View of Views" is a special case of View, where the type of each entry is itself a View. It is possible to make this, but before you try, please see below.
 
-6.2.3.1 You probably don't want this
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+You probably don't want this
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
romintomasetti, line 107: This one had the wrong underlining.

romintomasetti: image

dalg24 (commented):
Agreed that  we should not number by hand but I think we should preserve the numbering

docs/source/programmingguide.rst
```diff
@@ -2,6 +2,7 @@ Programming Guide
 #################
 
 .. toctree::
+   :numbered:
```
romintomasetti, line 5: @dalg24 I find the output satisfactory. Give it a try :wink:

PR #659, opened 2025-04-10
Aid in msvc building

OlekRaymond:
There was confusion by me and someone else who raised an issue trying to use `nvcc_wrapper` on windows. 

I think the wording of the `nvcc_wrapper` section in `Compiling.md` give the wrong impression. I have changed said wording (though I think it could be better) and added a note in `building_from_source.rst` so that what to do on windows can be seen easier.

I would also suggest adding something to the tutorials for windows/MSVC compilation. My current solution is not portable.

Thanks

### DIFF
--- a/docs/source/get-started/building-from-source.rst
+++ b/docs/source/get-started/building-from-source.rst
@@ -157,7 +157,7 @@ These options are generally useful for any CMake project:
   * ``SYCL``: Intel GPUs
     
   Example: ``-DKokkos_ENABLE_CUDA=ON``
-  Note that ``-DKokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE=ON`` is required when building with cuda and MSVC/CL
+  Note that ``-DKokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE=ON`` is required when building with CUDA and MSVC on Windows.
 
 
 
--- a/docs/source/get-started/building-from-source.rst
+++ b/docs/source/get-started/building-from-source.rst
@@ -157,7 +157,7 @@ These options are generally useful for any CMake project:
   * ``SYCL``: Intel GPUs
     
   Example: ``-DKokkos_ENABLE_CUDA=ON``
-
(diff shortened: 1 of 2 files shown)


docs/source/get-started/building-from-source.rst
```diff
@@ -157,7 +157,7 @@ These options are generally useful for any CMake project:
   * ``SYCL``: Intel GPUs
     
   Example: ``-DKokkos_ENABLE_CUDA=ON``
-
+  Note that ``-DKokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE=ON`` is required when building with cuda and MSVC/CL
```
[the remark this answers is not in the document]
masterleinad: The compiler command that is used by MSVC, see https://learn.microsoft.com/en-us/cpp/build/reference/compiler-command-line-syntax?view=msvc-170.

dalg24, line 160:
```suggestion
  Note that ``-DKokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE=ON`` is required when building with CUDA on Windows.
```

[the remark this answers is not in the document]
dalg24: I had not seen your comments.  Do you approve my changes?

OlekRaymond: I was under the impression that gcc on Windows would work, I haven't tried it so will revert back to your wisdom in this case.

dalg24: I don;t have access to Windows nor am familiar with their development environment but it is not clear to me that the nvcc_wrapper script would work on Windows and you definitely can't just use g++ w/o either cvcc_wrapper or the cmake as language option

OlekRaymond: So windows can't run shell scripts so nvcc_wrapper won't work, it uses batch, a different programming language for command line scripts.

You can use `g++` on windows but most people don't, the more common choice is clang because it can be downloaded in visual studio as an optional compiler. 

If you are using MSVC you are on windows, it isn't supported on Linux.

PR #666, opened 2025-04-22
Remove the `cppkokkos` domain

### LINKED ISSUE #559: consider removing cppkokkos domain
Currently we have forked the the sphinx cpp domain in a domain called cppkokkos. The idea is we wanted some specific markup for deprecated types and functions. However, we currently have a PR https://github.com/kokkos/kokkos-core-wiki/pull/451 that no one can work on for updating it to sync it with the latest sphinx release. This represents a maintenance burden for us, and will continue to cause problems in the future.

Furthermore, we can render deprecated code markings fine via C++ annotation syntax:
deprecated_doc_kokkos

If this is sufficient to render (and we could hack the CSS if we wanted to change how it looked), we could entirely drop cppkokkos and reduce our maintenance burden.

Are there any other reasons why we should keep the cppkokkos domain? @fnrizzi

nmm0:
This also replaces `kokkosinlinefunction` with `function`, and adds `KOKKOS*_FUNCTION` macros as function attributes so we can include them directly into the function signature.

Fixes #559


docs/source/conf.py
```diff
@@ -88,10 +87,14 @@
...
 
-rst_prolog = """
-.. role:: cppkokkos(code)
-    :language: cppkokkos
+cpp_id_attributes = [
+    'KOKKOS_FUNCTION',
+    'KOKKOS_INLINE_FUNCTION',
+    'KOKKOS_FORCEINLINE_FUNCTION',
+    'KOKKOS_DEFAULTED_FUNCTION'
+]
```
dalg24, line 95:
```suggestion
cpp_id_attributes = [
    'KOKKOS_FUNCTION',
    'KOKKOS_INLINE_FUNCTION',
    'KOKKOS_FORCEINLINE_FUNCTION',
    'KOKKOS_DEFAULTED_FUNCTION',
    'KOKKOS_RELOCATABLE_FUNCTION',
    'KOKKOS_DEDUCTION_GUIDE'
```

dalg24 (approved):

dalg24:
Very nice.  Thank you.

PR #667, opened 2025-04-22
refactor view documentation

nmm0:
This fixes a few issues and makes View a cross-referenceable entity. Also it brings everything in line stylistically with how we want to write documentation (i.e. use of cross-references and sphinx entities)

### DIFF
--- a/docs/source/API/core/view/view.rst
+++ b/docs/source/API/core/view/view.rst
@@ -53,9 +53,9 @@ Header File: ``<Kokkos_Core.hpp>``
    
    :tparam MemorySpace: Controls the storage location of the View.
 
-      If omitted the default memory space of the default execution space is used (i.e. :cpp:expr:`Kokkos::DefaultExecutionSpace::memory_space`)
+      If omitted the default memory space of the default execution space is used (i.e. :cpp:expr:`DefaultExecutionSpace::memory_space`)
 
-   :tparam MemoryTraits: Sets access properties via enum parameters for the templated :cpp:expr:`Kokkos::MemoryTraits<>` class.
+   :tparam MemoryTraits: Sets access properties via enum parameters for the templated :cpp:class:`MemoryTraits\<>` class.
 
       Possible template parameters are bit-combinations of the following flags:
 
@@ -65,7 +65,7 @@ Header File: ``<Kokkos_Core.hpp>``
          If the view is also :cpp:`const` this will trigger special load operations on GPUs (i.e. texture fetches).
       - ``Restrict``: There is no aliasing of the view by other data structures in the current scope.
 
-      Example usage: ``Kokkos::MemoryTraits<Kokkos::Unmanaged | Kokkos::RandomAccess>``
+      Example usage: :cpp:`MemoryTraits<Unmanaged | RandomAccess>`
 
    .. rubric:: Public Constants:
    
@@ -589,7 +589,7 @@ Header File: ``<Kokkos_Core.hpp>``
 
       :returns: an mdspan with extents and a layout converted from the :cpp:class:`View`'s *natural mdspan*.
 
-   .. cpp:function:: template <class OtherAccessorType = Kokkos::default_accessor<typename traits::value_type>> constexpr auto to_mdspan(const OtherAccessorType& other_accessor = OtherAccessorType{})
+   .. cpp:function:: template <class OtherAccessorType = default_accessor<typename traits::value_type>> constexpr auto to_mdspan(const OtherAccessorType& other_accessor = OtherAccessorType{})
 
       :tparam OtherAccessor: the target mdspan accessor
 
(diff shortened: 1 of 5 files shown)


docs/source/API/core/view/view.rst
```diff
@@ -3,580 +3,626 @@
...
+
+
+   .. rubric:: Spaces:
+
+   .. cpp:type:: execution_space
+
+      The :ref:`execution space <api-execution-spaces>` associated with the view, will be used for
+      performing view initialization, and certain deep_copy operations.
 
-* ``rank``: rank of the view (i.e. the dimensionality) **(until Kokkos 4.1)**
```
nmm0, line 98: I moved these guys to futher down

nmm0:
The page is too long for sreenshots but you can download the latest build artifact to view: https://github.com/kokkos/kokkos-core-wiki/actions/runs/14737077093/artifacts/3031708882

docs/source/API/core/view/view.rst
```diff
@@ -3,580 +3,626 @@
...
 
-.. cpp:type:: reference_type
+      Standard allocating constructor. The initialization is executed on the default
+      instance of the execution space corresponding to :cpp:type:`memory_space` and fences it.
 
-   return type of the view access operators.
+      :tparam IntType: an integral type
 
-.. cpp:type:: pointer_type
+      :param name: a user provided label, which is used for profiling and debugging purposes. Names are not required to be unique,
```
ldh4, line 215:
```suggestion
      :param name: a user provided label, which is used for profiling and debugging purposes. Names are not required to be unique.
```

docs/source/API/core/view/view.rst
```diff
@@ -3,580 +3,626 @@
...
-.. cpp:type:: array_layout
+   .. cpp:function:: View( const std::string& name, const array_layout& layout)
 
-   The Layout of the View.
+      Standard allocating constructor. The initialization is executed on the default
+      instance of the execution space corresponding to :cpp:type:`memory_space` and fences it.
 
-.. cpp:type:: size_type
+      :param name: a user provided label, which is used for profiling and debugging purposes.
+         Names are not required to be unique,
```
ldh4, line 231:
```suggestion
         Names are not required to be unique.
```

docs/source/templates/class_api.rst
```diff
@@ -145,7 +141,7 @@ Non-Member Functions
 
   :tparam ViewDst: the other
 
-  :return: true if :cpp:type:`View::value_type`, :cpp:type:`View::array_layout`, :cpp:type:`View::memory_space`, :cpp:member:`View::rank`, :cpp:func:`View::data()` and :cpp:expr:`View::extent(r)`, for :cpp:expr:`0<=r<rank`, match.
+  :return: true if :cpp:type:`View::value_type`, :cpp:type:`View::array_layout`, :cpp:type:`View::memory_space`, :cpp:member:`View::func`, :cpp:func:`View::data()` and :cpp:expr:`View::extent(r)`, for :cpp:expr:`0<=r<rank`, match.
```
ldh4, line 144: What is `View::func` referring to?

nmm0: That's a typo, meant to be:
```rst
:cpp:func:`View::rank`
```

I'll fix

nmm0:
After the discussion with @dalg24 and @crtrott  this morning, I managed to move all view members out-of-line. This means we can keep them categorized in sections in the ToC like they were before. (and they are a bit less indented).

Sphinx has a "namespace-push" directive that probably more accurately should be called "scope-push". It doesn't actually create a namespace entity but just says that everything between it and "namespace-pop" is part of that scope. I used that so all references should still work, i.e. `:cpp:func:\`View::rank\`` correctly references the view rank accessor.

Since everything is out of line I added a new heading "Class Interface" to differentiate between class members and the free functions.

crtrott:
Ok I think let's merge it and then incrementally improve where necessary.

crtrott (approved):

science-enthusiast:
Hi, Is there a reason why `traits` (an alias for `ViewTraits<...>`) is not explicitly mentioned in the documentation? I think it will be useful to discuss about the flags `is_managed` and `is_random_access` in the Memory Traits page and cross-reference it to a page which discusses other View Traits.

nmm0:
> Hi, Is there a reason why `traits` (an alias for `ViewTraits<...>`) is not explicitly mentioned in the documentation? I think it will be useful to discuss about the flags `is_managed` and `is_random_access` in the Memory Traits page and cross-reference it to a page which discusses the various View Traits.

@science-enthusiast I think it would make sense if you need this information for your PR to add it

science-enthusiast:
> @science-enthusiast I think it would make sense if you need this information for your PR to add it

I think it will be useful to have this information about `is_managed` and `is_random_access`. These flags are accessible to the user. But there are more aspects to View Traits. Maybe we could create page for it, so that the page for View doesn't get too long.

PR #668, opened 2025-04-23
DualView: Be more explicit about synchronization behavior for resize

masterleinad:
@JBludau found the current documentation for `DualView::resize` not clear enough in https://github.com/kokkos/kokkos/pull/8011#discussion_r2056021224. Thus, this pull request adds another advice for calling `sync` after `resize`.

### DIFF
--- a/docs/source/API/containers/DualView.rst
+++ b/docs/source/API/containers/DualView.rst
@@ -251,7 +251,7 @@ Description
     .. cppkokkos:function:: void resize(const size_t n0 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n1 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n2 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n3 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n4 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n5 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n6 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n7 = KOKKOS_IMPL_CTOR_DEFAULT_ARG);
 
        Resize both views, copying old contents into new if necessary. This method only copies the old
-       contents into the new View objects for the device which was last marked as modified. Thus, users are advised to call ``sync()`` before using the resized object.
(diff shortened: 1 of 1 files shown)


docs/source/API/containers/DualView.rst
```diff
@@ -251,7 +251,7 @@ Description
     .. cppkokkos:function:: void resize(const size_t n0 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n1 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n2 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n3 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n4 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n5 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n6 = KOKKOS_IMPL_CTOR_DEFAULT_ARG, const size_t n7 = KOKKOS_IMPL_CTOR_DEFAULT_ARG);
 
        Resize both views, copying old contents into new if necessary. This method only copies the old
-       contents into the new View objects for the device which was last marked as modified.
+       contents into the new View objects for the device which was last marked as modified. Thus, users are adviced to call ``sync()`` before using the resized object.
```
nmm0, line 254: Is this a recommendation, or basically a requirement?

JBludau, line 254:
```suggestion
       contents into the new View objects for the device which was last marked as modified. Thus, users are advised to call ``sync()`` before using the resized object.
```

masterleinad: If you are just calling `Kokkos::resize` it's not obvious to know whether the object was last modified on the host or the device (apart from calling `needs_sync_[host|device]`) and thus hard to know what the state after `resize` is. The flags aren't really in an undefined state but the underlying `Views` aren't synchronized so it's a recommendation to just call `sync` before using the object rather than relying on some assumed state. `sync` is pretty much a no-op if the requested side is already updated anyway.

JBludau: Then maybe we should express it as a requirement. It people don't do it, it is on them but in general we require it to be called after resize.

masterleinad: Done

masterleinad: Fair enough. I could see it both ways.

JBludau (approved):

PR #671, opened 2025-04-24
add python instructions to README

nmm0:

PR #672, opened 2025-04-24
take 2: add python version requirements and instrucitons to README

nmm0:
Supersedes https://github.com/kokkos/kokkos-core-wiki/pull/671 because I accidentally made a PR not from my fork :shrug:

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ Welcome to the Kokkos Documentation repository.  This is the source for https://
 
 ## Requirements to build html page locally
 
-The documentation requires Python 3.12 to build. Usually this can be installed with a system package manager, but if your system does not support python 3.12 you can install it easily with [pyenv](https://github.com/pyenv/pyenv).
+The documentation requires a minimum of Python 3.12 to build. Usually this can be installed with a system package manager, but if your system does not support python 3.12 or above you can install it easily with [pyenv](https://github.com/pyenv/pyenv).
 This is needed just for local render of documentation, so it can be checked before push.
 Requirements are in `build_requirements.txt`
 Could be installed with: `pip install -r build_requirements.txt`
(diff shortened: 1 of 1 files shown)


romintomasetti (changes requested):
I am really unconvinced by the necessity to explain how one can get `Python`.

How the developer should get `Python` is not part of this repository. To me, only the minimum `Python` version would be worth writing here.

README.md
```diff
@@ -3,10 +3,30 @@ Welcome to the Kokkos Documentation repository.  This is the source for https://
 
 ## Requirements to build html page locally
 
+The documentation requires a minimum of Python 3.12 to build. Usually this can be installed with a system package manager, but if your system does not support python 3.12 or above you can install it easily with [pyenv](https://github.com/pyenv/pyenv).
 This is needed just for local render of documentation, so it can be checked before push.
 Requirements are in `build_requirements.txt`
 Could be installed with: `pip install -r build_requirements.txt`
 
+We recommend using a virtual environment, e.g. if your system python is >= 3.12:
```
science-enthusiast, line 11: A minor point. It is better to capitalized Python when it is referred to as a proper noun. Lines 6 and 11.

dalg24:
No opinion on how to get Python.  I think that the suggestion to use venv with a short snippet the people can just copy and paste is a good addition though,

nmm0:
> I am really unconvinced by the necessity to explain how one can get `Python`.
> 
> How the developer should get `Python` is not part of this repository. To me, only the minimum `Python` version would be worth writing here.

I'd agree if this was more outwards facing. I added this instruction mainly for the benefit of others in the Kokkos team.

README.md
crtrott:
```suggestion
We recommend using a virtual environment, e.g. if your system Python is >= 3.12:
```

PR #676, opened 2025-04-28
Document deprecation of Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK in 4.7

dalg24:
Proposed updated doc for https://github.com/kokkos/kokkos/pull/8021

### DIFF
--- a/docs/source/get-started/configuration-guide.rst
+++ b/docs/source/get-started/configuration-guide.rst
@@ -182,9 +182,12 @@ Debugging
       * Use bounds checking - will increase runtime
       * ``OFF``
 
-    * * ``Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK``
+    * * ``Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK`` :red:`[Deprecated since 4.7]`
       * Debug check on dual views
-      * ``ON`` if ``CMAKE_BUILD_TYPE`` is ``Debug``, ``OFF`` otherwise
+      * (see below)
+
+
+``Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK`` default value is "``ON`` if ``CMAKE_BUILD_TYPE`` is ``Debug``, ``OFF`` otherwise" until 4.7 and effectively always ``ON`` since 4.7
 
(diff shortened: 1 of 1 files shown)


docs/source/get-started/configuration-guide.rst
```diff
@@ -182,9 +182,12 @@ Debugging
       * Use bounds checking - will increase runtime
       * ``OFF``
 
-    * * ``Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK``
+    * * ``Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK`` :red:`[Deprecated since 4.7]`
```
nmm0, line 185: Does the [deprecated](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) directive work here? I think it would be better as it is more canonical.

dalg24: It may but I chose to go for consistency with the rest of the page.
I would suggest we attempt to use the deprecated directive to all in a follow up PR.

nmm0: Ok fair enough!

nmm0 (approved):

docs/source/get-started/configuration-guide.rst
```diff
@@ -182,9 +182,12 @@ Debugging
...
       * Debug check on dual views
-      * ``ON`` if ``CMAKE_BUILD_TYPE`` is ``Debug``, ``OFF`` otherwise
+      * (see below)
+
+
+``Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK`` default value is "``ON`` if ``CMAKE_BUILD_TYPE`` is ``Debug``, ``OFF`` otherwise" until 4.7 and effectively always ``ON`` since 4.7
```
dalg24, line 190:
```suggestion
``Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK`` default value is "``ON`` if ``CMAKE_BUILD_TYPE`` is ``Debug``, ``OFF`` otherwise" until 4.7 and always ``ON`` since 4.7
```

PR #685, opened 2025-05-27
fix the number of arguments of `required_allocation_size`

tretre91:
In the documentation, `required_allocation_size` takes 9 arguments, even though Views are up to 8D (and DynRankView is up to 7D). The function actually takes 8 arguments (the last arg shouldn't be used for the DynRankView version)


masterleinad (commented):
`DynRankView` also has `1` as default argument. I'm not sure if that's intentional.

dalg24 (approved):

Issue #690, opened 2025-06-04
Missing reducers


nliber:
While writing the documentation to address https://github.com/kokkos/kokkos/issues/8097, we are missing documentation for the following reducers:

MaxFirstLoc
MaxFirstLocCustomComparator
MinFirstLoc
MinFirstLocCustomComparator
MinMaxFirstLastLoc
MinMaxFirstLastLocCustomComparator
FirstLoc
LastLoc
StdIsPartitioned
StdPartitionPoint

nliber:
Not sure yet if I'll address this with a separate PR to address https://github.com/kokkos/kokkos/issues/8097 or do it all as one.

PR #692, opened 2025-06-11
Adding Mxx*FirstLoc reducers

nliber:
This is to address https://github.com/kokkos/kokkos/issues/8097 and partially address #690.

Added new documentation for `MaxFirstLoc`, `MinFirstLoc` and `MinMaxFirstLastLoc`.

This documentation is a bit cleaner and more complete (includes defaulted `HostSpace`, `references_scalar()` and CTAD) than the existing `MaxLoc`, etc..  After this stabilizes, I'll update `MaxLoc`, etc. (and probably others) under a separate PR.

Added notes to `MaxLoc`, etc. about those who want a specific index for duplicate equivalent values should use `MaxFirstLoc`, etc. instead.

This note is under the description of `join`.

### DIFF
--- a/docs/source/API/core/builtinreducers/MinFirstLoc.rst
+++ b/docs/source/API/core/builtinreducers/MinFirstLoc.rst
@@ -76,7 +76,7 @@ Interface
    .. cpp:function:: void join(value_type& dest, const value_type& src) const;
 
       If ``src.val == dest.val`` && ``src.loc < dest.loc`` then ``dest.loc = src.loc``;
-      otherwise if ``src.val < dest.val`` then ``dest.val = src.val`` && `dest.loc = src.loc``.
+      otherwise if ``src.val < dest.val`` then ``dest.val = src.val`` && ``dest.loc = src.loc``.
 
    .. cpp:function:: void init(value_type& val) const;
 


docs/source/API/alphabetical.rst
```diff
@@ -150,6 +150,8 @@ Core
...
 +--------------------------------------------------------------------------------------+---------------------------+------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------+
+| `MaxFirstLoc <core/builtinreducers/MaxFirstLoc.html>`_                               | `Core <core-index.html>`_ | `Built-in Reducers <core/builtin_reducers.html>`_                      | Reducer for Reduction providing maximum and its first index position                                                                    |
++--------------------------------------------------------------------------------------+---------------------------+------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------+
```
science-enthusiast, line 154: Here and for `MinFirstLoc` instead of first index position we could say smallest index, like we do in the other places.

nliber: Fixed.

Good catch.  I hadn't come up with that term until I was in the middle of working on this.

PR #703, opened 2025-08-01
Allow view in unordered map

JBludau:
Adds documentation for https://github.com/kokkos/kokkos/pull/8182

### DIFF
--- a/docs/source/API/containers/Unordered-Map.rst
+++ b/docs/source/API/containers/Unordered-Map.rst
@@ -27,7 +27,7 @@ Description
 
    :tparam Key: Must be a POD (Plain Old Data type)
 
-   :tparam Value: `void` indicates an unordered set. If the map is created with the `SequentialHostInit` property, Views as `Value` are allowed. Otherwise the `Value` must be trivially copyable
+   :tparam Value: `void` indicates an unordered set. Otherwise the `Value` must be trivially copyable. (since Kokkos 4.7) If the map is created with the `SequentialHostInit` property, Views as `Value` are allowed.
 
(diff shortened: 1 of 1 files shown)


docs/source/API/containers/Unordered-Map.rst
```diff
@@ -39,6 +39,12 @@ Description
 
       .. warning:: Host Only
 
+   .. cpp:function:: UnorderedMap(const ALLOC_PROP &prop, uint32_t capacity_hint);
+
+      Create map using the properties with enough space for at least capacity_hint number of objects
+
+      .. warning:: Host Only
+
```
nmm0, line 47: Can you add documentation for alloc props? Maybe adding something about SequentialHostInit adding allowing the map to store views.

If this was added in 4.7 should do something like:

```suggestion
   .. cpp:function:: UnorderedMap(const ALLOC_PROP &prop, uint32_t capacity_hint);

      Create map using the properties with enough space for at least capacity_hint number of objects

      .. warning:: Host Only

      .. versionadded:: 4.7
```

nmm0 (approved):

docs/source/API/containers/Unordered-Map.rst
```diff
@@ -39,6 +42,13 @@ Description
...
+      .. versionadded:: 4.7
```
JBludau, line 51: how about:
```suggestion
      .. versionadded:: 4.2
      
      .. versionchanged:: 4.7
              :cpp:any:`prop` can now contain :cpp:class:`SequentialHostInit`
```

masterleinad: Sure

masterleinad (approved):

PR #723, opened 2025-08-29
Document `KOKKOS_IF_ON_{HOST,DEVICE}` macros

dalg24:


JBludau (approved):
I like it

docs/source/API/core/macros-special/if_on_host_or_device.rst
```diff
@@ -0,0 +1,111 @@
...
+function is running on the host (CPU) or a device (GPU, etc.). These macros
+provide an alternative to non-portable preprocessor idioms like #ifdef
+``__CUDA_ARCH__``.
+
+Motivation
+==========
+Traditional preprocessor directives like ``#ifdef __CUDA_ARCH__`` rely on a
+split compilation model, where host and device code are compiled in separate
+passes.  This model is not supported by modern unified compilation approaches,
+such as the one used by the NVIDIA HPC compiler (NVC++). As a result, code
```
crtrott, line 25: We need to get rid of this, NVC++ does not support CUDA anymore and they are not committed to doing it. Maybe mention SYCL or OpenMPTarget instead? How does it work there?

crtrott (commented):
Other than that NVC++ looks good.

PR #733, opened 2025-10-01
Add documentation fof gfx1201 architecture

Rombur:


docs/source/get-started/configuration-guide.rst
```diff
@@ -737,6 +737,11 @@ Kokkos will attempt to autodetect the architecture flag at configuration time.
       * MI50, MI60
       * (since Kokkos 4.2)
 
+    * * ``Kokkos_ARCH_AMD_GFX1201``
```
dalg24, line 740: Is that arch a more recent one?  I thought we were going newer on top

Rombur: It is on top of the consumer cards...

PR #740, opened 2025-10-15
Depracate array_layout in favor of layout_type

janciesko:


docs/source/API/core/view/layoutRight.rst
```diff
@@ -25,6 +25,10 @@ Description
 
    .. cpp:type:: array_layout
 
+       A tag signifying that this models the Layout concept. (Deprecated since Kokkos 5.0)
```
cwpearson:
```suggestion
       A tag signifying that this models the Layout concept. (Deprecated since Kokkos 5.0: use ``layout_type``)
```

PR #752, opened 2025-11-07
Update docs about array-valued reduction

brian-kelley:
Update API reference and programming guide example about using array-valued reductions. Previously these made it sound like reduce still worked with dyanamic-sized arrays, but this was removed a while back in favor of using Views (kokkos/kokkos/pull/4756). Also explicitly say that the result view must be rank 1 (now enforced in kokkos/kokkos/pull/8656).

**Question for clarification:** is parallel_reduce always synchronous if the result is a host view or array? I think the answer is yes but wanted to check since I explicitly mention that in the example with a HostSpace result view.

### DIFF
--- a/docs/source/ProgrammingGuide/ParallelDispatch.md
+++ b/docs/source/ProgrammingGuide/ParallelDispatch.md
@@ -223,7 +223,7 @@ parallel_reduce (X.extent(0), cs, sums);
 ```
 
 The result view could also use `Kokkos::HostSpace`, in which case
-the results accessing the results on the host requires a fence:
+accessing the results on the host requires a fence:
 
 ```c++
 Kokkos::View<float*, Kokkos::HostSpace> sums ("sums", numCols);


docs/source/ProgrammingGuide/ParallelDispatch.md
```diff
@@ -209,18 +209,34 @@ struct ColumnSums {
...
 
 View<float**> X ("X", numRows, numCols);
 // ... fill X before the following ...
 ColumnSums cs (X);
+Kokkos::View<float*> sums ("sums", numCols);
+parallel_reduce (X.extent(0), cs, sums);
+```
+
+The result view could also use `Kokkos::HostSpace`, in which case
+the results accessing the results on the host requires a fence:
```
masterleinad:
```suggestion
The result view could also use `Kokkos::HostSpace`, in which case
accessing the results on the host requires a fence:
```

PR #755, opened 2025-11-14
Set clang cuda version to 15.0.0 in the requirements

tretre91:
The minimum clang cuda version should be 15.0.0 according to what is set in kokkos_compiler_id.cmake (https://github.com/kokkos/kokkos/blob/33aa9d8ca3720873ca1aec19b828035b17db3395/cmake/kokkos_compiler_id.cmake#L158)


dalg24 (approved):
I think we promised 16 but in practice we have been tested 15

PR #835, opened 2026-03-20
Clarify that we require value types to be assignable in deep_copy

masterleinad:
As discussed in https://github.com/kokkos/kokkos/pull/8322#issuecomment-3188977990 and https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1773963510956809, `deep_copy` allows different value types as long as the destination value type is assignable from the source value type.


crtrott (changes requested):
I made an alternative change in #845 which I think is more accurate. Please take a look.

PR #847, opened 2026-04-09
Add some missing archs

JBludau:
Discovered in spack, see https://github.com/spack/spack-packages/pull/4161#discussion_r3052658941

I opted to not include BGQ since Power7 already is super old and BGQ is very specific to supercomputers (unlikely that someone has such a machine still)

### DIFF
--- a/docs/source/get-started/configuration-guide.rst
+++ b/docs/source/get-started/configuration-guide.rst
@@ -537,6 +537,10 @@ If cross-compiling, or if you want to be specific, the CPU architecture can be p
       - Xeon 9470C @ ANL Aurora
         Xeon @ LANL Crossroads
 
+    * - ``Kokkos_ARCH_ICX``
+      - Icelake
+      -
+
     * - ``Kokkos_ARCH_SKX``
       - Skylake/x86-64
       - 6130 @ OSU Pete
--- a/docs/source/get-started/configuration-guide.rst
+++ b/docs/source/get-started/configuration-guide.rst
@@ -538,7 +538,7 @@ If cross-compiling, or if you want to be specific, the CPU architecture can be p
         Xeon @ LANL Crossroads
 
     * - ``Kokkos_ARCH_ICX``
-      - Icelake
+      - Icelake/x86-64
       -
 
     * - ``Kokkos_ARCH_SKX``


docs/source/get-started/configuration-guide.rst
```diff
@@ -537,6 +541,10 @@ If cross-compiling, or if you want to be specific, the CPU architecture can be p
       - Xeon 9470C @ ANL Aurora
         Xeon @ LANL Crossroads
 
+    * - ``Kokkos_ARCH_ICX``
+      - Icelake
```
dalg24:
```suggestion
      - Icelake/x86-64
```
are these in inverse chronological order?

docs/source/get-started/configuration-guide.rst
```diff
@@ -523,6 +523,10 @@ If cross-compiling, or if you want to be specific, the CPU architecture can be p
       - Power8/Power ISA
       -
 
+    * - ``Kokkos_ARCH_POWER7``
+      - Power7/Power ISA
+      - (until Kokkos 4.2)
+
```
dalg24: I think we decided to remove that one, on the ground that this is an ancient architecture and that we did not expect they were much still deployed in the wild.
You wanted it for completeness?

JBludau: It was mainly brought up in https://github.com/spack/spack-packages/pull/4124#discussion_r3040036431
I personally would have gone the other way (removing it in both) but the consensus was to keep it in spack.
But we have another spack pr that puts version conditions on archs.

We could also say we don't care about the wiki and spack being different decide to not add it to the configuration-guide

docs/source/get-started/configuration-guide.rst
JBludau: I eyeballed where it would go assuming the first arch was the newest. My eyeballing seems to match my google but maybe I am bad at both of these

docs/source/get-started/configuration-guide.rst
```diff
@@ -523,6 +523,10 @@ If cross-compiling, or if you want to be specific, the CPU architecture can be p
       - Power8/Power ISA
       -
 
+    * - ``Kokkos_ARCH_POWER7``
+      - Power7/Power ISA
+      - (until Kokkos 4.2)
```
tpadioleau:
```suggestion
      - (removed in Kokkos 4.3)
```

To be consistent with the rest of the page ?

tpadioleau (approved):

docs/source/get-started/configuration-guide.rst
dalg24: Please drop the Power7 option

JBludau: done

dalg24 (approved):

PR #852, opened 2026-04-15
Document the use of pre-commit

dalg24:
@weinbe2
Is this helpful?

Regarding https://github.com/kokkos/kokkos/pull/9012#issuecomment-4247813898, I am not sure what went wrong when you apply the old formatting script.  But in our experience, pre-commit has been a very efficient way to make sure we all agree on the same clang-format version.


weinbe2:
@dalg24 yes, this is very helpful---sort of unfortunate that it seems like I need to use python to get it going, but the virtual environment (today I learned) makes it much more palatable. Thanks for putting this together!

masterleinad (approved):

PR #853, opened 2026-04-22
Don't specify a minimum compiler version for AppleClang for Kokkos 5

masterleinad:
As discussed in https://kokkosteam.slack.com/archives/G5CBLMFLP/p1775829992691849, we didn't update the minimum requirement from 4. This pull request suggests to just drop the compiler name from the list since it should largely behave like the respective `Clang` version (using `libc++`).

### DIFF
--- a/docs/source/get-started/requirements.rst
+++ b/docs/source/get-started/requirements.rst
@@ -30,9 +30,6 @@ Kokkos 5.x
     * * Clang (CUDA)
       * 15.0.0 and CUDA 11.8.0 (newer versions depending on LLVM support)
 
-    * * AppleClang 
-      * 8.0
-
     * * IntelLLVM (CPU)
       * 2022.0.0
 


tretre91 (approved):
Should we add a comment or a note to indicate that the supported AppleClang version should correspond to the one using the lowest llvm clang we support (possibly with a link to https://en.wikipedia.org/wiki/Xcode#Toolchain_versions)?

masterleinad:
> Should we add a comment or a note to indicate that the supported AppleClang version should correspond to the one using the lowest llvm clang we support (possibly with a link to [en.wikipedia.org/wiki/Xcode#Toolchain_versions](https://en.wikipedia.org/wiki/Xcode#Toolchain_versions))?

Not sure if that's worth it. We are saying
> Furthermore, compilers not listed here may work too.

already.

tretre91:
Ok, sounds good to me then

PR #866, opened 2026-06-02
Tutorials: updating the link to the slides

lucbv:
We are now automatically updating our slides and deploying them from the kokkos-tutorials pages on github.


masterleinad (commented):
That means that the YouTube link and the slides might don't correspond to each other anymore. Do we care?

lucbv:
It is not clear, we might want to serve the current slides on another page but then it will potentially be weird to serve two sets of slides?

PR #869, opened 2026-06-04
Add missing gfx 1151 and 1152 architectures

JBludau:


ldh4 (approved):
Fine as is, but maybe this list should be reordered to be in ascending gfx architecture ID order.

crtrott (approved):

JBludau:
> Fine as is, but maybe this list should be reordered to be in ascending gfx architecture ID order.

I think for now we went with: professional grade gpus first, then enthusiast grade

PR #872, opened 2026-06-09
Docs chapter numbering fix.

Quntized:


docs/source/ProgrammingGuide/ParallelDispatch.md
```diff
@@ -36,22 +36,21 @@ The entire parallel operation (for, reduce, or scan) shares the same instance of
...
 
-By default, lambdas capture nothing (as the default capture specifier `[]` specifies). This is not likely to be useful, since [`parallel_for()`](../API/core/parallel-dispatch/parallel_for) generally works by its side effects. Because Kokkos reserves the right to make copies of the closure, and its operations are potentially asynchronous users must ``capture by value'' to be semantically correct. We recommend doing so via the KOKKOS_LAMBDA macro for the outermost level of parallelism (see [Chapter 8](HierarchicalParallelism)).
+By default, lambdas capture nothing (as the default capture specifier `[]` specifies). This is not likely to be useful, since [`parallel_for()`](../API/core/parallel-dispatch/parallel_for) generally works by its side effects. Because Kokkos reserves the right to make copies of the closure, and its operations are potentially asynchronous users must ``capture by value`` to be semantically correct. We recommend doing so via the KOKKOS_LAMBDA macro for the outermost level of parallelism (see [Chapter 7](HierarchicalParallelism)).
```
Adrien-Tab:
```suggestion
By default, lambdas capture nothing (as the default capture specifier `[]` specifies). This is not likely to be useful, since [`parallel_for()`](../API/core/parallel-dispatch/parallel_for) generally works by its side effects. Because Kokkos reserves the right to make copies of the closure, and its operations are potentially asynchronous users must "capture by value" to be semantically correct. We recommend doing so via the KOKKOS_LAMBDA macro for the outermost level of parallelism (see [Chapter 7](HierarchicalParallelism)).
```

Adrien-Tab (approved):

masterleinad (commented):
I would just remove chapter numbers altogether. Having links is good enough.

Quntized:
@masterleinad That makes perfect sense. I will go ahead and replace all the hardcoded chapter and section which are mismatched with their actual titles

masterleinad (approved):

PR #874, opened 2026-06-10
StaticBatchSize and section on perf optimization

science-enthusiast:
* Documentation about StaticBatchSize.
* A section in ProgrammingGuide that can be used to collect all such opitimizations in one place.


docs/source/ProgrammingGuide/FurtherPerfOptimization.rst
```diff
@@ -0,0 +1,17 @@
+Further Performance Optimization
+================================
+
+In this chapter, we will discuss further ways to optimize performance. Different ideas involving varied settings have been collected here. We would like to highlight that these parameters should be tuned only to obtain extra performance gains, by exploiting problem specific structure and machine architecture. The values of these parameters need to be re-visited, when the user code is run in a new architecture. All these parameters have default values that don't enable the associated optimization.
+
+Static Batch Size
+-----------------
+
+For now, this parameter concerns `RangePolicy <../API/core/policies/RangePolicy.html>`__ based execution of :cpp:`Kokkos::parallel_for` in CUDA and HIP backends. `StaticBatchSize <../API/core/Execution-Policies.html#other-arguments-for-certain-execution-policies>`__ is a compile-time template parameter of the execution policy and it is of scalar integer type.
+
```
yasahi-hpc, line 13: Cay you add a code snippet here


```suggestion
For example, you can apply loop unrolling by 4 in the following manner.

.. code-block:: cpp

   Kokkos::parallel_for(
           Kokkos::RangePolicy<ExecSpace, Kokkos::Experimental::StaticBatchSize<4>>(0, N),
           KOKKOS_LAMBDA(const int i) {...});
```

docs/source/ProgrammingGuide/FurtherPerfOptimization.rst
```diff
@@ -0,0 +1,17 @@
...
+In GPU backends, like CUDA and HIP, :cpp:`Kokkos::parallel_for` executes the one-dimensional iteration space specified by the :cpp:`Kokkos::RangePolicy`, using a `grid-stride loop <https://developer.nvidia.com/blog/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/>`_. With grid-stride loops, iteration spaces of any sizes can be handled. As shown in the figure, each thread handles multiple elements, by taking steps that are equal to the size of the thread grid. By unrolling, we can make each thread process more elements, involving another smaller loop, before taking the bigger step. In the figure, we can see how a static batch size of 2 is handled. Thus, the same iteration space is processed by a smaller number of threads, where each thread does more work. While the same amount of memory traffic is involved in both the cases, unrolling leads to better utilization of the computational resources on the GPU.
```
yasahi-hpc:
```suggestion
In GPU backends, like CUDA and HIP, :cpp:`Kokkos::parallel_for` executes the one-dimensional iteration space specified by the :cpp:`Kokkos::RangePolicy`, using a `grid-stride loop <https://developer.nvidia.com/blog/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/>`_. With grid-stride loops, iteration spaces of any sizes can be handled. As shown in the figure, each thread handles multiple elements, by taking steps that are equal to the size of the thread grid. By unrolling, we can make each thread process more elements, involving another smaller loop, before taking the bigger step. In the figure, we can see how a static batch size of 2 is handled. Thus, the same iteration space is processed by a smaller number of threads, where each thread does more work. While the same amount of memory traffic is involved in both the cases, unrolling leads to better utilization of the computational resources on the GPU. This is particularly important to saturate the modern GPUs by incrementing the instruction level parallelism.
```

Better to explain when it matters.

docs/source/ProgrammingGuide/FurtherPerfOptimization.rst
```diff
@@ -0,0 +1,17 @@
+Further Performance Optimization
+================================
+
+In this chapter, we will discuss further ways to optimize performance. Different ideas involving varied settings have been collected here. We would like to highlight that these parameters should be tuned only to obtain extra performance gains, by exploiting problem specific structure and machine architecture. The values of these parameters need to be re-visited, when the user code is run in a new architecture. All these parameters have default values that don't enable the associated optimization.
+
+Static Batch Size
+-----------------
+
+For now, this parameter concerns `RangePolicy <../API/core/policies/RangePolicy.html>`__ based execution of :cpp:`Kokkos::parallel_for` in CUDA and HIP backends. `StaticBatchSize <../API/core/Execution-Policies.html#other-arguments-for-certain-execution-policies>`__ is a compile-time template parameter of the execution policy and it is of scalar integer type.
```
yasahi-hpc: Explain that this is an experimental feature for the moment. Also, can you confirm whether it is not implemented for SYCL?

```suggestion
For now, this parameter concerns `RangePolicy <../API/core/policies/RangePolicy.html>`__ based execution of :cpp:`Kokkos::parallel_for` in CUDA and HIP backends. `StaticBatchSize <../API/core/Execution-Policies.html#other-arguments-for-certain-execution-policies>`__ is a compile-time template parameter of the execution policy and it is of scalar integer type. This functionality is offered as an Experimental feature.
```

yasahi-hpc (changes requested):

docs/source/ProgrammingGuide/FurtherPerfOptimization.rst
science-enthusiast: I have come across PRs for CUDA and HIP. I will check whether it has been implemented for SYCL.

science-enthusiast: StaticBatchSize has not yet been implemented for SYCL backend.

docs/source/ProgrammingGuide/FurtherPerfOptimization.rst
```diff
@@ -0,0 +1,28 @@
...
+
+In GPU backends, like CUDA and HIP, :cpp:`Kokkos::parallel_for` executes the one-dimensional iteration space specified by the :cpp:`Kokkos::RangePolicy`, using a `grid-stride loop <https://developer.nvidia.com/blog/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/>`_. With a grid-stride loop, iteration spaces of any sizes can be handled. As shown in the figure, each thread handles multiple elements, by taking steps that are the size of the thread grid. By unrolling, we can make each thread process more elements, involving another smaller loop, before taking the bigger step. In the figure, we can see that for a static batch size of 2. Thus, the same iteration space is processed by a smaller number of threads, where each thread does more work. While the same amount of memory traffic is involved in both the cases, unrolling leads to better utilization of the computational resources on the GPU. This is particularly helpful for saturating modern GPUs by increasing the instruction level parallelism.
+
+|node|
```
yasahi-hpc: Do we need this?
```suggestion

```

yasahi-hpc (approved):
Looks good

docs/source/ProgrammingGuide/FurtherPerfOptimization.rst
science-enthusiast: I think the figure helps to understand how the unrolling happens.

docs/source/ProgrammingGuide/FurtherPerfOptimization.rst
yasahi-hpc: Sure

docs/source/ProgrammingGuide/FurtherPerfOptimization.rst
yasahi-hpc: OK, I understand. 

Probably better to rename the tag and add a caption.

`|node|` -> `|unroll| Figure 1.1 Grid-stride loop with and without unrolling`

science-enthusiast: Yes, the name has to be changed. I overlooked it when I copy pasted the syntax from elsewhere.

docs/source/ProgrammingGuide/FurtherPerfOptimization.rst
```diff
@@ -0,0 +1,28 @@
+Further Performance Optimization
+================================
+
+.. |sbs| image:: figures/grid_stride_loop_static_batch.png
+   :alt: Figure 17.1 One-dimensional grid-stride loop, without and with unrolling
+
+In this chapter, we will discuss further ways to optimize performance. Different ideas involving varied settings have been collected here. We would like to highlight that these parameters should be tuned only to obtain extra performance gains, by exploiting problem specific structure and machine architecture. The values of these parameters need to be re-visited, when the user code is run in a new architecture. All these parameters have default values that don't enable the associated optimization.
```
masterleinad:
```suggestion
In this chapter, we will discuss further ways to optimize performance. Different ideas involving varied settings have been collected here. We would like to highlight that these parameters should be tuned only to obtain extra performance gains, by exploiting problem-specific structure and machine architecture. The values of these parameters need to be re-visited, when the user code is run on a new architecture. All these parameters have default values that don't enable the associated optimization.
```

docs/source/ProgrammingGuide/FurtherPerfOptimization.rst
```diff
@@ -0,0 +1,28 @@
...
+
+In GPU backends, like CUDA and HIP, :cpp:`Kokkos::parallel_for` executes the one-dimensional iteration space specified by the :cpp:`Kokkos::RangePolicy`, using a `grid-stride loop <https://developer.nvidia.com/blog/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/>`_. With a grid-stride loop, iteration spaces of any sizes can be handled. As shown in the figure, each thread handles multiple elements, by taking steps that are the size of the thread grid. By unrolling, we can make each thread process more elements, involving another smaller loop, before taking the bigger step. In the figure, we can see that for a static batch size of 2. Thus, the same iteration space is processed by a smaller number of threads, where each thread does more work. While the same amount of memory traffic is involved in both the cases, unrolling leads to better utilization of the computational resources on the GPU. This is particularly helpful for saturating modern GPUs by increasing the instruction level parallelism.
```
masterleinad:
```suggestion
In GPU backends, like CUDA and HIP, :cpp:`Kokkos::parallel_for` executes the one-dimensional iteration space specified by the :cpp:`Kokkos::RangePolicy`, using a `grid-stride loop <https://developer.nvidia.com/blog/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/>`_. With a grid-stride loop, iteration spaces of any sizes can be handled. As shown in the figure, each thread handles multiple elements, by taking steps that are the size of the thread grid. Using ``StaticBatchSize``, we can make each thread process more elements, involving another smaller loop, before taking the bigger step. In the figure, we can see that for a batch size of 2. Thus, the same iteration space is processed by a smaller number of threads, where each thread does more work. While the same amount of memory traffic is involved in both cases, this increases the instruction level parallelism which might lead to better utilization of the computational resources on the GPU.
```

docs/source/ProgrammingGuide/FurtherPerfOptimization.rst
```diff
@@ -0,0 +1,28 @@
...
+For example, you can apply an unrolling factor of 4 in the following manner.
+
+.. code-block:: cpp
+
+  Kokkos::parallel_for(
+      Kokkos::RangePolicy<
+          Kokkos::Experimental::StaticBatchSize<4> /*, other arguments */>(N),
+      KOKKOS_LAMBDA(const int i){...});
+
+We have to keep in mind that grid-stride loop unrolling doesn't to lead data being shared between the unrolled indices. Thus, for problems with minimal or no data re-use like the Stream benchmark, one can obtain speed-up with static batch size > 1.
```
masterleinad: I don't think there are many applications that actually benefit from reusing data between work items for a 1D range loop. I think the point is rather to use it for kernels that don't saturate memory bandwidth and have very little compute.

```suggestion
```

masterleinad (commented):
Can you move `LaunchBounds` to the new section as well?

docs/source/API/core/Execution-Policies.rst
```diff
@@ -84,14 +84,38 @@ Execution Policies generally accept compile time arguments via template paramete
...
-    * * LaunchBounds
-      * ``LaunchBounds<MaxThreads, MinBlocks>``
-      * Specifies hints to to the compiler about CUDA/HIP launch bounds.
-
     * * WorkTag
       * ``SomeClass``
-      * Specify the work tag type used to call the functor operator. Can be any arbitrary tag type (i.e. an [empty](https://en.cppreference.com/w/cpp/types/is_empty) struct or class). Defaults to ``void``.
+      * Specify the work tag type used to call the functor operator. Can be any arbitrary tag type (i.e. an `empty <https://en.cppreference.com/w/cpp/types/is_empty>`_ struct or class). Defaults to ``void``.
+
+Other arguments for certain Execution Policies
```
masterleinad, line 95: Ah, my bad! I misread this caption to read for "certian backends" instead of for "certain Execution Policies". Please move `LaunchBounds` back.

science-enthusiast: I don't know the internals of Team Policies yet. From a grep based search, I see LaunchBounds only concerning RangePolicy and MDRangePolicy. Does LaunchBounds have an effect on Team Policies also?

crtrott: I believe it should.

science-enthusiast: Thank you for your clarification. I will move `LaunchBounds` back to its earlier place, where we mention that it is specific to CUDA and HIP.

crtrott:
I have some concerns about effectively documenting and thus guaranteeing what we do internally for our implementation. I am not sure we should. Or at least we should make it VERY explicit that this is not semantically guaranteed behavior and can change at any time. Code which relies on this knowledge is invalid Kokkos code.

crtrott (changes requested):
Blocking. We need to discuss the implied guarantee in how this internally works.

science-enthusiast:
I agree that we should be clear about what is semantically guaranteed and we should not let user code depend on internal details.

PR #877, opened 2026-06-26
Add extents method

yasahi-hpc:


crtrott:
I think there are a bunch more things missing which are now coming from mdspan. like extents_type is not even defined. You wanna take that on or hand off?

yasahi-hpc:
Fair enough
I will do that

yasahi-hpc:
Currently, I am considering to have two distinct pages for Basic and Legacy Views. 
What do you think?

masterleinad:
> Currently, I am considering to have two distinct pages for Basic and Legacy Views.
> What do you think?

There isn't really much of a point spending significant effort on documenting legacy View since it doesn't really exist ion the latest releases without using `IMPL` options.

yasahi-hpc:
> > Currently, I am considering to have two distinct pages for Basic and Legacy Views.
> > What do you think?
> 
> There isn't really much of a point spending significant effort on documenting legacy View since it doesn't really exist ion the latest releases without using `IMPL` options.

Yeah, I will rename current `view.rst` into `legacy_view.rst` and make a new `view.rst` based on `BasicView`. `legacy_view.rst` is referred from `view.rst` as a deprecated remark

docs/source/API/core/view/view.rst
```diff
@@ -477,6 +477,11 @@ Data Layout, Dimensions, Strides
 
    - :cpp:any:`dim` must be smaller than :cpp:func:`rank`.
 
+.. cpp:function:: constexpr extents_type extents() const
```
dalg24, line 480: Is extents_type defined anywhere?

yasahi-hpc: No
Actually, so many things are missing

docs/source/API/core/view/view.rst
```diff
@@ -477,6 +477,11 @@ Data Layout, Dimensions, Strides
...
+
+   :return: the extents of all dimensions
+   :since: since 5.0
```
dalg24: As far as I can tell, the `:since:` directive is something we made up.
https://www.sphinx-doc.org/en/master/usage/domains/cpp.html#info-field-lists

I can only find it in the Array doc, I would rather not spread that pattern.
Use [`:versionadded:`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded) instead

yasahi-hpc: I can replace :since: with [:versionadded:](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded) in the Array doc if needed

dalg24: please do

yasahi-hpc: Done #908

dalg24:
```suggestion
   :versionadded: 5.0
```

docs/source/API/core/view/view.rst
```diff
@@ -477,6 +477,11 @@ Data Layout, Dimensions, Strides
 
    - :cpp:any:`dim` must be smaller than :cpp:func:`rank`.
 
+.. cpp:function:: constexpr extents_type extents() const
+
+   :return: the extents of all dimensions
+   :versionadded: 5.0
```
dalg24:
```suggestion
   .. versionadded:: 5.0
```

yasahi-hpc:
This should be addressed by #907
Can I close this PR? @dalg24

dalg24:
> This should be addressed by #907 Can I close this PR? @dalg24

Sure

PR #886, opened 2026-07-28
Document removal of DEPRECATED_CODE_4 option and more

dalg24:
Corresponds to kokkos/kokkos#9386

Note that I did not attempt to document aliases when they were missing.  I am not planning to do it but feel free to push if you want to do it.

### DIFF
--- a/docs/source/API/containers/DynRankView.rst
+++ b/docs/source/API/containers/DynRankView.rst
@@ -193,6 +193,8 @@ Description
 
       A specialization tag used for partial specialization of the mapping construct underlying a Kokkos ``DynRankView``.
 
+      .. versionremoved:: 5.3
+
    .. rubric:: Constructors
 
    .. cpp:function:: DynRankView()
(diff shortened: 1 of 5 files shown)


JBludau (commented):
Do you also want to update https://kokkos.org/kokkos-core-wiki/deprecations.html?

dalg24:
> Do you also want to update https://kokkos.org/kokkos-core-wiki/deprecations.html?

You mean add notes about version removed?
I don't feel strongly about this but just noting that we have not done so before.

JBludau:
> > Do you also want to update https://kokkos.org/kokkos-core-wiki/deprecations.html?
> 
> You mean add notes about version removed? I don't feel strongly about this but just noting that we have not done so before.

I missed that the deprecation of `HostMirror` is already in the list.
Not sure if we should start a removal list ... I think we don't need to

JBludau (approved):

PR #887, opened 2026-07-29
Prefer versionadded directive for create_mirror_view_and_copy

dalg24:


masterleinad (commented):
What's the difference in rendering? We are using "since Kokkos" in more places than "versionadded".

dalg24:
> What's the difference in rendering? We are using "since Kokkos" in more places than "versionadded".

Example here https://kokkos.org/kokkos-core-wiki/API/core/view/view.html#_CPPv4N4View16host_mirror_typeE

We want the sphinx directive

JBludau (approved):

masterleinad (approved):
Fine with me but it seems to make more sense to do that in a larger scope for consistency.

dalg24:
> Fine with me but it seems to make more sense to do that in a larger scope for consistency.

Sure.  I was specifically looking over a few entries on the 5.2 changelog.

PR #891, opened 2026-07-30
Improve documentation of abort

dalg24:
Prompted by a question on Slack

> I think this is implied but not explicitly stated, and I wanted to check: does void Kokkos::abort(const char *const msg) assume that msg is a fixed string, i.e., not a printf-style format string?

I went ahead and modernize the page to use more idiomatic restructured text directives.
I also added an example.

### DIFF
--- a/docs/source/API/core/utilities/abort.rst
+++ b/docs/source/API/core/utilities/abort.rst
@@ -45,6 +45,13 @@ Backend-Specific Behavior
    the SYCL backend and ``NDEBUG`` is defined, the function does **not** cause
    abnormal termination. Instead, it prints to the standard output stream and
    continues program execution.
+   See :ref:`Known Issues <known-issues-sycl-abort>`.
+
+.. warning::
+   **NextSilicon Backend:** When calling :cpp:func:`abort` from a parallel
+   region with the NextSilicon backend, the function will cause abnormal
+   program termination but will not print the message if the region has been
+   offloaded to the accelerator.
 
 
 Example
--- a/docs/source/API/core/utilities/abort.rst
+++ b/docs/source/API/core/utilities/abort.rst
@@ -13,7 +13,8 @@ Usage
 
     Kokkos::abort("helpful error message");
 
-Causes abnormal program termination and prints an error message.
+Causes abnormal program termination and prints an error message to the standard
+error stream.
 This function can be called from both host and device code, including within
 parallel kernels.
 
@@ -23,7 +24,8 @@ Interface
 
 .. cpp:function:: KOKKOS_FUNCTION void abort(const char * msg);
 
-   :param msg: Null-terminated string containing the error message to print before termination
+   :param msg: Null-terminated string containing the error message to print
+     before aborting the process.
    :returns: Does not return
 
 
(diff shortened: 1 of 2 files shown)


dalg24:
@kc9jud please review

docs/source/API/core/utilities/abort.rst
```diff
@@ -1,22 +1,75 @@
...
+* The fine-grained header ``<Kokkos_Abort.hpp>`` was added in version 4.2
+
+Backend-Specific Behavior
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. warning::
+   **SYCL Backend:** When calling :cpp:func:`abort` from a parallel region with
+   the SYCL backend and ``NDEBUG`` is defined, the function does **not** cause
+   abnormal termination. Instead, it prints to the standard output stream and
+   continues program execution.
```
dalg24, line 47: @masterleinad here is what I wanted feedback from you on

masterleinad: Yes, the abort mechanism relies on `NDEBUG` not being defined which is typically only the case in `Debug` mode; `RelWithDebInfo` anf `Release` define `NDEBUG`. It might be worth adding that.
```suggestion
.. warning::
   **SYCL Backend:** When calling :cpp:func:`abort` from a parallel region with
   the SYCL backend and ``NDEBUG`` is defined, the function does **not** cause
   abnormal termination. Instead, it prints to the standard output stream and
   continues program execution. Note that ``NDEBUG`` is typically defined for ``CMake``'s ``RelWithDebInfo`` and ``Release`` build types, but not for ``Debug``.
```

docs/source/API/core/utilities/abort.rst
```diff
@@ -1,22 +1,75 @@
...
+
```
kc9jud, line 55:
```
.. warning::
   **NextSilicon Backend:** When calling :cpp:func:`abort` from a parallel region with
   the NextSilicon backend, the function will cause program termination but will not
   print the message if the region has been offloaded to the accelerator.

```

kc9jud (dismissed):
Here's some suggested language documenting the current limitations of the NextSilicon backend.

docs/source/API/core/utilities/abort.rst
dalg24: I added a cross-reference to the SYCL section in Known Issues.
Would you edit that entry instead?  I don't want the notes sections to be too long and I find it nice to have one place that recalls all known issues.

docs/source/known-issues.rst
```diff
@@ -139,6 +139,10 @@ SYCL
     struct sycl::is_device_copyable<MyComparator>
       : std::true_type {};
 
+.. _known-issues-sycl-abort:
+
+- Longer explanation here explaining limitations with :cpp:func:`abort` in the SYCL backend.
```
dalg24: @masterleinad please edit this instead

dalg24: ping

masterleinad:
```suggestion
- When calling :cpp:func:`abort` from a parallel region with the SYCL backend
  and ``NDEBUG`` is defined, the function does **not** cause abnormal
  termination. Instead, it prints to the standard output stream and continues
  program execution. Note that ``NDEBUG`` is typically defined for ``CMake``'s
  ``RelWithDebInfo`` and ``Release`` build types, but not for ``Debug``.
```

masterleinad (approved):

PR #892, opened 2026-07-30
 Document new public Kokkos_InitializeFinalize.hpp header 

masterleinad:
Follow-up to https://github.com/kokkos/kokkos/pull/9309.

### DIFF
--- a/docs/source/API/core/initialize_finalize/InitializationSettings.rst
+++ b/docs/source/API/core/initialize_finalize/InitializationSettings.rst
@@ -4,7 +4,7 @@
 .. role:: cpp(code)
    :language: cpp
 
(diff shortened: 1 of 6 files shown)


dalg24 (commented):
I am curious what others think, whether we want a note saying that before 5.3, users are expected to include `<Kokkos_Core.hpp>`, or whether that is clear enough as is.

dalg24:
> I am curious what others think, whether we want a note saying that before 5.3, users are expected to include `<Kokkos_Core.hpp>`, or whether that is clear enough as is.

Per discussion on the developer channel, this is good enough until someone complains.

dalg24 (approved):

PR #897, opened 2026-08-06
Document RC branch creation in the release process

dalg24:
Populate further our release process.
Here detailing how the maintainer should create the release candidate branch.

### DIFF
--- a/docs/source/developer-guides/release-process.rst
+++ b/docs/source/developer-guides/release-process.rst
@@ -3,6 +3,93 @@ Release Process
 
 (Work In Progress)
 
+Create Release Branch and Update Project Version
+-------------------------------------------------
+
+.. note::
+   **Version Numbering Scheme**
+
+   Development versions use ``.99`` as the patch number to ensure they always
+   compare as more recent than any release in that series (e.g., ``4.2.99`` >
+   ``4.2.5``). This allows code to reliably discriminate between development
+   and release versions.
+
+.. important::
+   Steps 1-7 must be completed in sequence without merging other changes to
+   ``develop``. This ensures accurate version tracking throughout the codebase.
+
+1. Check that the ``develop`` branch is in decent shape with the Continuous
+   Integration Working Group, that all nightly builds are passing, and that there
+   are no outstanding issues in the integration testing.
+
+2. Create the release candidate branch:
+
+.. code-block:: console
+
+   git checkout -b release-candidate-X.Y.0
+
+3. Update the version number to ``X.Y.0`` in the root ``CMakeLists.txt``:
+
+.. code-block:: cmake
+
+   # Edit these lines in CMakeLists.txt:
+   set(Kokkos_VERSION_MAJOR X)
+   set(Kokkos_VERSION_MINOR Y)
+   set(Kokkos_VERSION_PATCH 0)
+
+Then commit the change:
+
+.. code-block:: console
+
+   git commit -s -m 'Set version number to X.Y.0' CMakeLists.txt
+
+4. Push the release candidate branch to the upstream repository:
+
+.. code-block:: console
+
+   git push https://github.com/kokkos/kokkos.git release-candidate-X.Y.0
+
+5. Create and checkout a new ``bump_version_number`` branch with ``develop`` as
+   the starting point:
+
+.. code-block:: console
+
+   git checkout -b bump_version_number develop
+
+6. Update the version from ``X.(Y-1).99`` to ``X.Y.99`` in the root ``CMakeLists.txt``:
+
+.. code-block:: cmake
+
+   # Edit these lines in CMakeLists.txt:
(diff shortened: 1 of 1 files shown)


docs/source/developer-guides/release-process.rst
```diff
@@ -3,6 +3,93 @@ Release Process
...
+-------------------------------------------------
+
+.. note::
+   **Version Numbering Scheme**
+
+   Development versions use ``.99`` as the patch number to ensure they always
+   compare as more recent than any release in that series (e.g., ``4.2.99`` >
+   ``4.2.5``). This allows code to reliably discriminate between development
+   and release versions.
+
```
dalg24, line 17: Not sure whether we want this note.
The `.99` patch version number is discussed here https://kokkos.org/kokkos-core-wiki/API/core/Macros.html#version-macros
That said I don't know that there is any other documentation for it.
Let me know what you think.

dalg24:
I was planning to discuss patch releases in the next PR.
Let me know if you want it as part of this one.

lucbv (commented):
I would add a comment somewhere to mention that this guide refers to major and minor version release. 
Actually I am not sure it works completely for major release because of the deprecation flags and such?
You want to add something about coordinating with the rest of the Ecosystem that the rc is created and that the develop version has been bumped. KK uses that info as you know : )

dalg24:
> I would add a comment somewhere to mention that this guide refers to major and minor version release.

I was going to refine the language as I document patch releases in the next PR.  Is that OK with you?

> Actually I am not sure it works completely for major release because of the deprecation flags and such?

This is a good point but I would like to defer to when we actually document the support policy and deprecations process.

> You want to add something about coordinating with the rest of the Ecosystem that the rc is created and that the develop version has been bumped. KK uses that info as you know : )

I expected the coordination aspect will emerge when we start integrating the release coordinator checklist with preparation for the release briefing.

docs/source/developer-guides/release-process.rst
```diff
@@ -3,6 +3,93 @@ Release Process
...
+.. code-block:: console
+
+   git commit -s -m 'Set version number to X.Y.0' CMakeLists.txt
+
+4. Push the release candidate branch to the upstream repository:
+
+.. code-block:: console
+
+   git push https://github.com/kokkos/kokkos.git release-candidate-X.Y.0
```
masterleinad, line 51: This is the only step that requires special permissions (at least for Kokkos Core). Since the "Final Tasks" section explicitly mentions "maintainers", it might be worth it to also clarify the audience for these instructions. If it's the release manager or any other "normal" developer, it's helpful to point out that this step needs reaching out to one of the maintainers.

docs/source/developer-guides/release-process.rst
```diff
@@ -3,6 +3,93 @@ Release Process
...
+
+   Development versions use ``.99`` as the patch number to ensure they always
+   compare as more recent than any release in that series (e.g., ``4.2.99`` >
+   ``4.2.5``). This allows code to reliably discriminate between development
+   and release versions.
+
+.. important::
+   Steps 1-7 must be completed in sequence without merging other changes to
+   ``develop``. This ensures accurate version tracking throughout the codebase.
+
```
ndellingwood, line 21: Would you consider including something like a comment or a sub-header specifying these are the steps for major and minor releases?

ndellingwood: A separate sub-section for patch release branching could be added after maybe, indicating that the most recent release (rather than develop) is used as the base for creating the patch release branch

docs/source/developer-guides/release-process.rst
ndellingwood: I like that you included this here since it consolidates info relevant to the versioning, without this I think tracking down the link you posted would add time and cognitive load if someone was seeking out clarification on the use of .99

docs/source/developer-guides/release-process.rst
```diff
@@ -3,6 +3,93 @@ Release Process
 
 (Work In Progress)
 
+Create Release Branch and Update Project Version
+-------------------------------------------------
```
dalg24, line 7:
```suggestion
-------------------------------------------------
(For maintainers)
```

docs/source/developer-guides/release-process.rst
dalg24: Clarified https://github.com/kokkos/kokkos-core-wiki/pull/897/changes#r3731276444

docs/source/developer-guides/release-process.rst
dalg24: My intent was to do patch release next.
Then, I will add a (leading) release timeline section.
I also expect we will integrate the release coordination (release briefing preparation).

lucbv (approved):

crtrott (approved):

PR #898, opened 2026-08-07
Document how to initiate patch releases and rules to cherry-pick changes into RC branches

dalg24:


docs/source/developer-guides/release-process.rst
```diff
@@ -91,6 +94,110 @@ to maintain version integrity in the development branch.
    branch has been created and that the version bump PR needs to be merged as
    the next change to ``develop``.
 
+Patch Releases (X.Y.Z)
+~~~~~~~~~~~~~~~~~~~~~~
+
+Patch releases are created from the previous release tag to incorporate
+critical bug fixes into an existing release series.
```
lucbv, line 101: I do not feel that every bug fix needs to rise to the level of "critical" to make it into a patch release.
Additionally, I like this quick explanation what we want to put in a patch release, it begs the question: why not have the same paragraph for minor version release?
```suggestion
Patch releases are created from the previous release tag to incorporate
bug fixes into an existing release series.
```

lucbv: I see that the paragraph for minor release content was actually added further below...

docs/source/developer-guides/release-process.rst
```diff
@@ -91,6 +94,110 @@ to maintain version integrity in the development branch.
...
+Changes should follow the develop-first workflow unless there is an exceptional
+reason, such as:
+
+- Develop has diverged significantly, making cherry-picking impractical
+- Develop is temporarily broken or in an untestable state
+- The bug only exists in the release branch due to subsequent fixes in develop
+- Develop contains incompatible changes requiring a fundamentally different fix
+
+1. **Merge to develop first:** All changes must be integrated and tested on the
+   ``develop`` branch before being considered for backporting.
```
lucbv, line 153:
```suggestion
   ``develop`` branch before being considered for cherry-picking.
```

docs/source/developer-guides/release-process.rst
```diff
@@ -91,6 +94,110 @@ to maintain version integrity in the development branch.
...
+
+2. **Get authorization:** Once merged to develop, obtain approval from a maintainer,
+   through the weekly developer meeting, or on the `#nucleus <https://kokkosteam.slack.com/archives/G5CBLMFLP>`_
+   channel before backporting.
```
lucbv, line 157:
```suggestion
   channel before cherry-picking.
```

docs/source/developer-guides/release-process.rst
```diff
@@ -91,6 +94,110 @@ to maintain version integrity in the development branch.
...
+3. **Open cherry-pick PR:** Create a pull request targeting the release candidate
+   branch with:
+
+   - **Title format:** ``[X.Y.Z] Original Well-Crafted Subject Line``
+   - **Description starts with:** "Cherry-picking the changes from PR #1234 into
+     the X.Y.(Z+1) release candidate branch"
+
+.. tip::
+   Developers are encouraged (but not required) to seek approval **before**
+   opening the backport PR to avoid unnecessary work if the change is deemed
```
lucbv, line 168:
```suggestion
   opening the cherry-pick PR to avoid unnecessary work if the change is deemed
```

lucbv (changes requested):
Small changes requested, mostly some stylistic change and maybe bringing slightly down the requirements to make it into a patch release.

docs/source/developer-guides/release-process.rst
dalg24: How strong do you feel on this.  I think "backporting" was right.

lucbv: I get what you are saying but I think if we emphasize back porting it goes against the idea that we do not want big back ports that require refactor, we would much rather get cherry pick because it is safer overall for patch-releases...

dalg24: > Backporting is the act of taking a fix or feature from a newer version of software and applying it to an older, stable version. Cherry-picking is the specific action of taking one single commit from a branch and applying it to another branch. Cherry-picking is the most common tool used to complete a backport.

(From the AI overview on my Google search)

In summary, cherry-picking is a technical tool; backporting is the goal.

lucbv: Okay, then I approve the PR if you replace "backporting" with "non refactor based backporting" : )
But in general I do not mind so do as you wish

lucbv (approved):

crtrott (approved):
Fine, though I do agree with Luc on the cherry-picking vs back porting. I feel back porting is something you do from some newer version that was already released to some older version. But whatever :-)

PR #901, opened 2026-08-12
Describe release timeline

dalg24:


lucbv (changes requested):
The release briefing is missing from the timeline

crtrott (approved):

lucbv (approved):
I think it's fine.
Note for future reference: we should state at the top of the file what the intent it, do we want to detail the technical step so someone can reproduce them or just have people understand the process and what to expect?

PR #903, opened 2026-08-12
parallel_scan.rst: no TeamPolicy version

cwpearson:
parallel_scan does not support TeamPolicy.


masterleinad (approved):
Shouldn't be very hard to do if we make it a priority, though.

PR #906, opened 2026-08-18
Document print_configuration

### LINKED ISSUE #888: `print_configuration` is not documented

dalg24:
Assisted-by: Claude:claude-4.5-sonnet

Close #888

### DIFF
--- a/docs/source/API/core/utilities/print_configuration.rst
+++ b/docs/source/API/core/utilities/print_configuration.rst
@@ -1,0 +1,92 @@
+``print_configuration``
+=======================
+
+Defined in header ``<Kokkos_Core.hpp>``
+
+Usage
+-----
+
+.. code-block:: cpp
+
+    Kokkos::print_configuration(std::cout);
+    Kokkos::print_configuration(output_stream, /*verbose=*/ true);
+
+Prints Kokkos configuration information ("bill of materials") to an output
+stream.
+This includes both compile-time configuration details (enabled backends,
+compiler settings, version information, build configuration) and runtime
+information determined during :cpp:func:`initialize` (such as number of
+threads for host parallel backends, or number of visible devices and device IDs
+for device backends).
+
+API Reference
+-------------
+
+.. cpp:function:: void print_configuration(std::ostream& os, bool verbose = false)
+
+   Prints Kokkos configuration information to the specified output stream.
+
+   :param os: Output stream to write configuration information to
+   :param verbose: If ``true``, prints additional detailed information
+
+
+Example
+-------
+
+.. code-block:: cpp
+
+   #include <Kokkos_Core.hpp>
+   #include <iostream>
+   
+   int main(int argc, char* argv[]) {
+       Kokkos::initialize(argc, argv);
+       
+       // Print basic configuration to the standard output
+       Kokkos::print_configuration(std::cout);
+       
+       // Write vebose configuration to a log file
+       std::ofstream log_file("kokkos_config.log");
+       if (log_file.is_open()) {
+           Kokkos::print_configuration(log_file, /*verbose=*/ true);
+           log_file.close();
+       }
+       
+       Kokkos::finalize();
+   }
+
+Notes
+-----
+
+.. note::
+   * Useful for debugging and verifying the Kokkos build configuration and runtime settings
+   * The output includes both compile-time and runtime information:
+
+     * **Compile-time**: Enabled execution spaces, memory spaces, compiler information, and feature flags
+     * **Runtime**: Number of threads (for host parallel backends like OpenMP, Threads), number of visible devices and device ID in use (for device backends like CUDA, HIP, SYCL)
+
+   * Must be called after :cpp:func:`Kokkos::initialize` to display runtime information
+
+.. warning::
+   Kokkos makes no guarantees about the format of the output. The format is subject to change between releases.
+
+.. tip::
+   You can print the configuration to standard output without adding an
+   explicit :cpp:func:`print_configuration` call and recompiling by setting the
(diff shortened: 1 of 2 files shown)


docs/source/API/core/utilities/print_configuration.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Notes
+-----
+
+.. note::
+   * Useful for debugging and verifying the Kokkos build configuration and runtime settings
+   * The output includes both compile-time and runtime information:
+
+     * **Compile-time**: Enabled execution spaces, memory spaces, compiler information, and feature flags
+     * **Runtime**: Number of threads (for host parallel backends like OpenMP, Threads), number of visible devices and device ID in use (for device backends like CUDA, HIP, SYCL)
```
JBludau: This is kind of a repetition of what was said at the top in the `Usage` section

docs/source/API/core/utilities/print_configuration.rst
```diff
@@ -0,0 +1,92 @@
...
+
+   * Must be called after :cpp:func:`Kokkos::initialize` to display runtime information
```
JBludau: should we promote that to a warning?

docs/source/API/core/utilities/print_configuration.rst
```diff
@@ -0,0 +1,92 @@
...
+       Kokkos::print_configuration(std::cout);
+       
+       // Write vebose configuration to a log file
+       std::ofstream log_file("kokkos_config.log");
+       if (log_file.is_open()) {
+           Kokkos::print_configuration(log_file, /*verbose=*/ true);
+           log_file.close();
+       }
+       
+       Kokkos::finalize();
```
JBludau, line 55: Should this also have the scope between initialize and finalize?

docs/source/API/core/utilities/print_configuration.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Usage
+-----
+
+.. code-block:: cpp
+
+    Kokkos::print_configuration(std::cout);
+    Kokkos::print_configuration(output_stream, /*verbose=*/ true);
+
+Prints Kokkos configuration information ("bill of materials") to an output
```
JBludau: I am not sure if `bill of materials` is a good thing to put here ... I would omit it

docs/source/API/core/utilities/print_configuration.rst
dalg24: I can add if you prefer.  There is no Kokkos object so it is not technically needed.
Let me know if you want me to add it.

docs/source/API/core/utilities/print_configuration.rst
dalg24: Why not?
That was the claimed motivation to add it
https://github.com/kokkos/kokkos/issues/632
and there is still a comment to that effect in the code
https://github.com/kokkos/kokkos/blob/dfa6599eb2d3277815cc2f406cc56cf66e02f85c/core/src/impl/Kokkos_RuntimeInfo.hpp#L11-L12

Is that misusing the terminology?
I thought it was on point, because it does spell out what Kokkos it is (version and configuration settings).

docs/source/API/core/utilities/print_configuration.rst
dalg24: Sure.
Make the one below two bulleted points or separate block?

docs/source/API/core/utilities/print_configuration.rst
dalg24: What change would you like to see?

JBludau: I would just remove it. The info does not have to be twice on the page

docs/source/API/core/utilities/print_configuration.rst
JBludau: no preference

docs/source/API/core/utilities/print_configuration.rst
JBludau: Maybe we should add it for consistency? but I do not feel strong about this

docs/source/API/core/utilities/print_configuration.rst
JBludau: I think it is too close to https://en.wikipedia.org/wiki/Software_supply_chain (SBOM)
I rather see it as a configuration of internal settings than a list of software that is used in Kokkos

docs/source/API/core/utilities/print_configuration.rst
dalg24: Which would you remove?  This one goes more into "detail".

JBludau: this one here. Since we do not give any promises about what it contains or how it is formatted, I would just leave it at "Useful for debugging and verifying the Kokkos build configuration"

docs/source/API/core/utilities/print_configuration.rst
dalg24: I dropped it.  On second thought that is true for all Kokkos functionality unless specified otherwise.

docs/source/API/core/utilities/print_configuration.rst
dalg24: I removed it.

docs/source/API/core/utilities/print_configuration.rst
dalg24: Done.

docs/source/API/core/utilities/print_configuration.rst
dalg24: Sure.  Done.

JBludau (approved):

